Issue #19, opened 2019-02-27
Add setInverval support to `gloo-timers`


fitzgen:
It currently supports `setTimeout` and `clearTimeout` but does not support `setInterval` and `clearInterval`.

OddCoincidence:
I'd like to work on this if it's not already claimed!

fitzgen:
This was fixed in #24

PR #24, opened 2019-03-14
Add setInterval support to `gloo_timers`

OddCoincidence:
Here's my initial attempt at #19. I'm pretty new to wasm{,-bindgen} and only have rudimentary experience with streams so apologies if anything is not right.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 /target
-/crates/*/target
 **/*.rs.bk
 Cargo.lock
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
+/crates/*/target
 **/*.rs.bk
 Cargo.lock
--- a/crates/timers/src/lib.rs
+++ b/crates/timers/src/lib.rs
@@ -322,19 +322,17 @@ impl Interval {
     /// ```no_run
     /// use gloo_timers::Interval;
     ///
-    /// let timeout = Interval::new(1_000, move || {
+    /// let interval = Interval::new(1_000, move || {
     ///     // Do something...
     /// });
     /// ```
     pub fn new<F>(millis: u32, callback: F) -> Interval
     where
-        F: 'static + FnOnce(),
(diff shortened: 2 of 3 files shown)


crates/timers/src/lib.rs
```diff
@@ -286,3 +287,201 @@ impl Future for TimeoutFuture {
...
+    /// ```no_run
+    /// use gloo_timers::Interval;
+    ///
+    /// let timeout = Interval::new(1_000, move || {
+    ///     // Do something...
+    /// });
+    /// ```
+    pub fn new<F>(millis: u32, callback: F) -> Interval
+    where
+        F: 'static + FnOnce(),
```
fitzgen: `FnOnce` is for closures that potentially consume values they close over and drop them. Therefore it is only safe to call them once, since callign them a second time will result in using already-dropped values. This makes sense for timers, which are invoked at most once.

Intervals on the other hand, are explicitly invoked many times at some frequency. Instead, we should use `FnMut`, which is a closure that is allowed to mutate its captured values, but not drop them.

https://doc.rust-lang.org/nightly/std/ops/trait.FnOnce.html
https://doc.rust-lang.org/nightly/std/ops/trait.FnMut.html
https://doc.rust-lang.org/nightly/book/ch13-01-closures.html

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
+/crates/*/target
```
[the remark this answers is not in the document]
OddCoincidence: I did yeah - I just tried to reproduce from a clean checkout though and didn't get it this time. The only difference is that I've updated to the latest stable toolchain since then. I'll remove this.

crates/timers/src/lib.rs
OddCoincidence: Done!

OddCoincidence:
@fitzgen I think I addressed your comments above, let me know if there's anything else that needs to be done for this!

fitzgen:
Looks great, thanks @OddCoincidence!

PR #26, opened 2019-03-16
Added router crate

David-OConnor:
Uses generics  as endpoints: requires a function to update app state, and one to update a wasm_bindgen Closure.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,5 +9,6 @@ version = "0.1.0"
 
 [dependencies]
 gloo-timers = { version = "0.1.0", path = "crates/timers" }
+gloo-router = { version = "0.1.0", path = "crates/router" }
 
 [workspace]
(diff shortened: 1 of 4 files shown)


David-OConnor:
Would apprec wisdom from anyone knows how to make the module doc check , or unwrap/expect_throw work.

fitzgen:
Hey David -- I'd like us not to merge PRs to gloo without review by at least one other maintainer. Additionally, I'd like to get to a place where we have largely already done design/consensus work up front, and agree on at least a skeleton of an approach and niche in the design space that we are aiming for. I'm thinking we should have proposals like https://github.com/rustwasm/gloo/blob/master/.github/ISSUE_TEMPLATE/propose_design.md and consensus in that proposal thread before we get to sending PRs.

With that in mind, I'm going to revert this PR merge, but I'd love to see a proposal for this design in the router tracking issue, with some discussion about the potential alternatives, prior art, etc. And then who knows, maybe we will merge this exactly as-is again! Or maybe it will go through a little design iteration and get even better? But I'd like this to be a community project where we can all contribute to designs.

Thanks for understanding!

PR #32, opened 2019-03-19
Add gloo-console-timer

### LINKED ISSUE #16: RAII type for console.time and console.timeEnd
Let's pull it out of https://rustwasm.github.io/book/game-of-life/time-profiling.html#time-each-universetick-with-consoletime-and-consoletimeend

cbrevik:
Resolves #16 

Adds the `gloo-console-timer` crate with the `ConsoleTimer` RAII type.

Docs feature the preferred `scope`-method for measurement:
```rust
let value = ConsoleTimer::scope("foo", || {
    // Place code to be measured here
});
```

Was a bit unsure how to write tests here, since it seems I cannot verify through tests that the measurement including label is printed to the console.


crates/console-timer/src/lib.rs
```diff
@@ -0,0 +1,75 @@
+/*!
+
+Measuring invocation time on the Web can be done with `console.time`
+and `console.timeEnd`.
```
fitzgen: Nitpick: can we link to MDN for more context here? https://developer.mozilla.org/en-US/docs/Web/API/console#Timers

And also mention that the timings end up in the browser's developer tools console and sometimes their performance view? I think we can crib a lot of this text from the rustwasm book: https://rustwasm.github.io/docs/book/reference/time-profiling.html#the-consoletime-and-consoletimeend-functions

crates/console-timer/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+
+This API wraps both the `time` and `timeEnd` calls into a single type
+named `ConsoleTimer`, ensuring both are called.
+
+## Scoped measurement
```
fitzgen: Can we also have a section on the RAII-style usage below this? Potentially showing timing asynchronous operations, where the scoped callbacks approach can't be used.

crates/console-timer/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "gloo-console-timer"
+version = "0.1.0"
+authors = ["Christian Brevik <christian.brevik@gmail.com>"]
```
fitzgen: Can you make this `"The Rust and WebAssembly Working Group"` like the root `Cargo.toml` does?

I just realized that the `crates/timers/Cargo.toml` doesn't use that same author line, and I'll file a PR to fix it.

cbrevik: Woops, didn't check the root `Cargo.toml`. Will fix!

cbrevik:
Adressed feedback, but I see one of the doc tests doesn't build, so will have to fix that first. (The one using `gloo_timers`).

cbrevik:
Maybe there is a better way to handle this than to `ignore` the test? 

I didn't want to add dependencies just to make that example compile.

crates/console-timer/src/lib.rs
```diff
@@ -32,7 +32,7 @@ For scenarios where `ConsoleTimer::scope` can't be used, like with
 asynchronous operations, you can use `ConsoleTimer::new` to create a timer.
 The measurement ends when the timer object goes out of scope / is dropped.
 
-```no_run
+```ignore
```
fitzgen: What about adding a `dev-dependency`? Exists for precisely this kind of thing ;)

cbrevik: Hehe alright, not clear on what best practices are in cases like these. Done! 👍

Issue #27, opened 2019-03-16
How deeply should we integrate with the ecosystem?


Pauan:
Right now web-sys provides very low-level access to the web APIs. My understanding is that Gloo is intended to create idiomatic high-level Rust APIs on top of those low-level APIs. However, there are different degrees of "high-level".

As an example, it's easy to wrap `setTimeout` and `setInterval` to make a nice convenient API that uses `FnOnce` and `FnMut`. But that's only slightly more high-level than using the APIs directly.

On the other hand, we could wrap `setTimeout` and `setInterval` into a `Future` and `Stream`, which is definitely much higher level (and more idiomatic for Rust).

So when we are faced with this choice between mid-level and high-level, what should we choose? I think we should have a consistent approach across the Gloo ecosystem.

I propose that we always expose the high-level APIs (`Future`, `Stream`, etc.) and then have a submodule which exposes the mid-level APIs.

That makes it clear that the high-level APIs are the idiomatic APIs that generally should be used, but still makes the mid-level APIs accessible for unusual circumstances.

To start the bikeshedding, I propose we consistently use the name `raw` for the mid-level submodule.

David-OConnor:
We could have dedicated utility crate that wraps common/verbose `web_sys` features, which the bulk of the modules (I agree they should be high-level) can call. Eg:

```rust
/// A convenience function for logging to the web browser's console.  See also
/// the log! macro, which is more flexible.
pub fn log<S: ToString>(text: S) {
    web_sys::console::log_1(&text.to_string().into());
}

/// Convenience function to avoid repeating expect logic.
pub fn window() -> web_sys::Window {
    web_sys::window().expect("Can't find the global Window")
}

/// Convenience function to access the web_sys DOM document.
pub fn document() -> web_sys::Document {
    window().document().expect("Can't find document")
}

/// Simplify getting the value of input elements; required due to the need to cast
/// from general nodes/elements to HTML__Elements.
pub fn input_value(target: &web_sys::EventTarget) -> String {
    if let Some(input) = target.dyn_ref::<web_sys::HtmlInputElement>() {
        return input.value();
    }
    if let Some(input) = target.dyn_ref::<web_sys::HtmlTextAreaElement>() {
        return input.value();
    }
    if let Some(input) = target.dyn_ref::<web_sys::HtmlSelectElement>() {
        return input.value();
    }
    "".into()
}

```

Pauan:
[...]

PR #34, opened 2019-03-20
Adding in note about how to handle Futures / Streams

Pauan:
See #27

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,6 +126,10 @@ Here is a checklist that all Gloo utility crates should fulfill:
       escape hatch for dropping down to raw `web_sys` bindings when an API isn't
       fully supported by the crate yet.
 
+* [ ] There is a loose hierarchy with "mid-level" APIs (which are essentially thin wrappers over the low-level APIs), and "high-level" APIs (which make more substantial changes).
+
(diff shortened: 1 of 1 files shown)


fitzgen:
Oh also: since `Future` and `Stream` are both defined in the `futures` crate, does it make sense to have separate submodules for them? Should they both be inside the crate's `futures` submodule?

fitzgen:
Oh also: we should mention that the `futures` usage should be behind a `"futures"` cargo feature!

fitzgen:
Also also: should the top-level `gloo` crate have a `futures` feature that enables all of the sub-crates' `futures` features?

I'm leaning towards "yes" for convenience, but I'm also interested in what others think.

Pauan:
> Oh also: since Future and Stream are both defined in the futures crate, does it make sense to have separate submodules for them? Should they both be inside the crate's futures submodule?

The `futures` (plural) crate has separate `future` and `stream` (singular) submodules, so I just went with that.

I don't have strong opinions one way or the other, but I do like segregating by types, and it also paves the way to other types (like `signal`).

> Also also: should the top-level gloo crate have a futures feature that enables all of the sub-crates' futures features?

That sounds good to me.

Pauan:
I added a note about high-level and mid-level APIs.

I think we need to have consensus on what the future/stream cargo feature should be called before I make a note about it.

PR #42, opened 2019-03-23
[WIP] gloo-events

### LINKED ISSUE #30: RAII system for handling event listeners
## Summary

I propose to implement an API which makes it easy to use Rust's RAII-based system for event listener registration/deregistration.

## Motivation

When registering an event listener, it is often needed to deregister it later.

In JavaScript this is done with the `removeEventListener` method, but that is verbose and error prone. It also does not clean up Rust resources (e.g. the memory for the closure).

## Detailed Explanation

```rust
pub struct EventListener<'a, F> where F: ?Sized {
    node: EventTarget,
    kind: &'a str,
    callback: Closure<F>,
}

impl<'a, F> EventListener<'a, F> where F: ?Sized + WasmClosure {
    #[inline]
    pub fn new<F>(node: &EventTarget, kind: &'a str, f: F) -> Self {
        let callback = Closure::wrap(Box::new(f) as Box<F>);

        node.add_event_listener_with_callback(kind, callback.as_ref().unchecked_ref()).unwrap();

        Self {
            node: node.clone(),
            kind,
            callback,
        }
    }
}

impl<'a, F> Drop for EventListener<'a, F> where F: ?Sized {
    #[inline]
    fn drop(&mut self) {
        self.node.remove_event_listener_with_callback(self.kind, self.callback.as_ref().unchecked_ref()).unwrap();
    }
}
```

The idea is that you can simply call `EventListener::new(&node, "foo", move |e| { ... })` and when the `EventListener` is dropped it will then cleanup the Rust memory for the closure and it will also deregister the event listener.

This is a general purpose idiomatic mid-level API which can work with *any* event on *any* DOM node. I expect it to be used pervasively both within Gloo and outside of Gloo.

For a mid-level API like this, there aren't really any alternatives. Some small things can be tweaked, and I haven't thought very hard about how to support `FnOnce` (but I'm sure it is possible). Ideas and refinements are welcome.

## Unresolved Questions

It is possible to build a higher level API on top of this.

This higher level API gives full static type support for JS events, making it impossible to get the types wrong.

The way that it works is that a new trait like this is created:

```rust
pub trait StaticEvent {
    const TAG: &'static str;
    type Event;
    fn wrap(event: Self::Event) -> Self;
}
```

And a new `fn` is created, like this:

```rust
pub fn on<A, F>(node: &EventTarget, f: F) -> EventListener<'static, FnMut(A::Event)>
    where A: StaticEvent,
          F: FnMut(A) + 'static {
    EventListener::new(node, A::TAG, move |e| {
        f(A::wrap(e));
    })
}
```

Now you can create various events like this:

```rust
struct ClickEvent {
    event: web_sys::MouseEvent,
}

impl ClickEvent {
    // Various methods specific to the `click` event
}

impl StaticEvent for ClickEvent {
    const TAG: &'static str = "click";

    type Event = web_sys::MouseEvent;

    fn wrap(event: Self::Event) -> Self {
        ClickEvent { event }
    }
}
```

```rust
struct MouseDownEvent {
    event: web_sys::MouseEvent,
}

impl MouseDownEvent {
    // Various methods specific to the `mousedown` event
}

impl StaticEvent for MouseDownEvent {
    const TAG: &'static str = "mousedown";

    type Event = web_sys::MouseEvent;

    fn wrap(event: Self::Event) -> Self {
        MouseDownEvent { event }
    }
}
```

And now *finally* we can actually use the API:

```rust
on(&node, move |e: ClickEvent| {
    ...
})

on(&node, move |e: MouseDownEvent| {
    ...
})
```

This API has some *huge* benefits:

* You no longer need to specify a string for the event type (so no more typos).

* You no longer need to do any coercions to the correct type (that's all handled automatically now).

* It *guarantees* that the static type is correct (in other words, if you have a `click` event, it is guaranteed to be matched to a `ClickEvent` Rust type, and vice versa).

This technique was pioneered by stdweb, and it works out really well in practice. It may also be necessary for solving https://github.com/rustwasm/wasm-bindgen/issues/1348

But that's something which can be layered on top of the mid-level `EventListener` API, so I defer it to later.

Pauan:
Do not merge yet!

Since the current plan is to create a high-level statically typed events system, what should we do with this mid-level API?

I think it makes sense for the high-level and mid-level API to both go into the `gloo-events` crate, but should we put `EventListener` into a submodule? Or is it okay to leave it at the root? Will that confuse users?

Fixes #30


Pauan:
Also, what name should we use for the string for the event type? Right now I call it `kind` (since `type` cannot be used), but perhaps `tag` would be better? Or maybe something more verbose like `event_type`?

Pauan:
Also, right now it only accepts `&str`. What if we want to allow for passing in a `String` into the `EventListener` (and have it be owned by the `EventListener`)? That can be useful for strings which are dynamically created.

crates/events/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use wasm_bindgen::{JsCast, UnwrapThrowExt};
+use wasm_bindgen::closure::Closure;
+use web_sys::{Event, EventTarget};
+
+
+///
+pub struct EventListener<'a> {
+    target: EventTarget,
+    event_type: &'a str,
```
OddCoincidence: When would the `event_type` not be `'static`?

Pauan: It allows you to dynamically choose the event type at runtime:

```rust
let s: String = get_string_somehow();

EventListener::new(&foo, &s, move |e| {
    // ...
})
```

Imagine `get_string_somehow()` returns a string at runtime (e.g. based on user input)

Although this is rather rare, it does happen!

As for `StaticEvent`... I'm not actually sure of any use cases right now, but since we cannot change it later, I figured it was best to make it more flexible (and consistent with `EventListener`).

If I can't think of any use cases, I'm fine with `StaticEvent` always using `'static` (though `EventListener` should still use `'a`)

coolreader18: A better example than user input would be from JS code calling into Rust.

Pauan: Sure, there's lots of examples where dynamic events are useful. I only listed one of them.

fitzgen:
> I think it makes sense for the high-level and mid-level API to both go into the `gloo-events` crate, but should we put `EventListener` into a submodule? Or is it okay to leave it at the root? Will that confuse users?

I think that is fine, since the high-level, static API doesn't handle dynamic event listening, so it makes sense to have both, IMO.

Pauan:
@fitzgen Before I can fix this up, I think we need to resolve these two issues:

https://github.com/rustwasm/gloo/issues/30#issuecomment-475866803
https://github.com/rustwasm/gloo/pull/42#issuecomment-475869524

fitzgen:
> Also, right now it only accepts `&str`. What if we want to allow for passing in a `String` into the `EventListener` (and have it be owned by the `EventListener`)? That can be useful for strings which are dynamically created.

Sounds like a perfect opportunity to use [`Cow<str>`](https://doc.rust-lang.org/nightly/std/borrow/enum.Cow.html)

Pauan:
@fitzgen I don't think that works, since `Cow` has a lifetime, which is attached to the lifetime of the original `String` (so it's not any better than just using `&'a str` directly).

The goal is to be able to dynamically create a `String`, and then shift ownership of the `String` into the `EventListener` (so that way the `String` will get cleaned up at the same time as the `EventListener`).

fitzgen:
> I don't think that works, since `Cow` has a lifetime, which is attached to the lifetime of the original `String` (so it's not any better than just using `&'a str` directly).

The lifetime can be `'static` (optionally; can also be `'a` for short-term listeners) which would enable usage like this:

```rust
// assuming this signature:
impl<'a> EventListener<'a> {
    pub fn new<S, F>(target: &EventTarget, event_type: S, callback: F) -> Self
    where
        S: Into<Cow<'a, str>>,
        F: FnMut(Event) + 'static,
    { ... }
}

// Enables both
EventListener::new(&target, "click", |event| { ... });

// and
let s: String = ...;
EventListener::new(&target, s, |event| { ... });
```

We could remove the `<'a>` and require `S: Into<Cow<'static, str>>`, since stack-lifetime event listeners aren't really a thing. Then the event type would be either a `&'static str` or a `String` (no `&'a str`s).

The other option would be adding the `S` type parameter to `EventListener` itself, but I think that is a bit much here: we don't need the monomorphization for perf reasons, and it would make the API that much more confusing, and compile times take that much longer.

Pauan:
@fitzgen Interesting, I hadn't seen that `Into<Cow<...>>` pattern before.

That does mean that all dynamic event types *must* be `String`, it wouldn't be possible to use `&'a str` anymore.

But as you say, allowing for `&'a str` is a little weird, since it's basically saying that the string is stack allocated but the `EventListener` wouldn't be.

Pauan:
Something new I just discovered:

https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners

It is apparently possible to significantly improve the performance of the browser by setting `passive: true` on the event listeners.

The only downside is that now you cannot call `event.preventDefault()`.

My thinking is that we should make `passive: true` the default, and provide a way to manually set `passive: false` if the user needs to call `event.preventDefault()`.

@fitzgen Any thoughts on this?

fitzgen:
I think that the mid-level API probably shouldn't be as opinionated as that, but maybe the higher-level API could be.

I think we want an `EventListener` constructor that takes a `web_sys::AddEventListenerOptions` reference either way, so that users can control all this.

Orthogonal to the expressive ability to define event listener options: As another layer, the `web_sys::AddEventListenerOptions` builder API isn't too bad, but I think we can do better than allocating the same, fresh objects (e.g. `{ passive: true }`) over and over every time we add event listeners. There are only three boolean options on the builder, so only eight different combinations. We could define every combination up front, store them in an array. They would be indexed by the binary number for whether each option was set or not. This way, we could re-use the JS objects, instead of re-allocating them every time we add an event listener.

Pauan:
> I think that the mid-level API probably shouldn't be as opinionated as that, but maybe the higher-level API could be.

To be clear, we're just discussing the defaults. Right now the browsers will default to `passive: true` for *some* events, but not others.

The reason the browsers have (started to) default to `true` is because 99+% of event listeners don't need to use `event.preventDefault()`, and if there is even a *single* event listener with `passive: false` it can significantly slow down *the entire web page*.

The only reason the browsers aren't *always* defaulting to `true` is because of backwards compatibility. But we aren't limited by backwards compatibility.

What you're suggesting is essentially that the easy way should incur significant performance costs (even if the event listener doesn't actually use `event.preventDefault()`), and the user has to go out of their way to use the fast version.

What I'm suggesting is that the easy way should be fast, and that the user should go out of their way to enable `event.preventDefault()` if and only if they actually need it. I think that is both more intuitive and fits in perfectly with Rust's overall philosophy of making performance costs explicit.

> I think we want an EventListener constructor that takes a web_sys::AddEventListenerOptions reference either way, so that users can control all this.

Agreed.

> As another layer, the web_sys::AddEventListenerOptions builder API isn't too bad, but I think we can do better than allocating the same, fresh objects (e.g. { passive: true }) over and over every time we add event listeners.

Also agreed.

fitzgen:
> What I'm suggesting is that the easy way should be fast, and that the user should go out of their way to enable `event.preventDefault()` if and only if they actually need it. I think that is both more intuitive and fits in perfectly with Rust's overall philosophy of making performance costs explicit.

Ok, I'm convinced.

Pauan:
@fitzgen Okay, great, I think that nails down all the design elements, so I'll try to get this finished ASAP, since I know it's blocking other work.

Pauan:
@fitzgen I fixed all the issues that were brought up.

I'm very happy with both the API and the implementation, but I want to make sure we have consensus that this is the right public API before I do the hard work of adding in docs and unit tests.

Implementation-wise, since `new` and `once` are going to be used 99%+ of the time, I created thread-locals for those, but not for the other 6 combinations of `EventListenerOptions` (this can be improved later, or users can manually cache if they wish).

crates/events/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+        options.once(self.once);
+        options.passive(self.passive);
+
+        options
+    }
+}
+
+// This defaults passive to true to avoid performance issues in browsers:
+// https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Improving_scrolling_performance_with_passive_listeners
+thread_local! {
```
fitzgen: Aside: We might want to use something other than `thread_local!` here; IIRC it has an unfortunate code size footprint over unsafe static refcells (assuming single-threaded wasm). This can be addressed in follow up PRs (and actually measured). We might want to make a `wasm_thread_local!` thing that wraps up the unsafety.

crates/events/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+pub struct EventListener {
+    target: EventTarget,
+    event_type: Cow<'static, str>,
+    callback: Option<Closure<FnMut(Event)>>,
+    is_capture: bool,
+}
+
+impl EventListener {
+    #[inline]
+    fn new_(
```
fitzgen: Nitpick: I'd name this `new_internal` instead of only appending an underscore.

crates/events/src/lib.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub once: bool,
+
+    ///
+    pub passive: bool,
+}
+
+impl EventListenerOptions {
+    ///
+    #[inline]
+    pub fn to_js(&self) -> AddEventListenerOptions {
```
fitzgen: `to_raw`?

crates/events/src/lib.rs
Pauan: I don't think we should assume single-threaded wasm, since multi-threaded wasm is just around the corner.

In any case, I agree we can investigate that later, since it's just an internal optimization.

crates/events/src/lib.rs
Pauan: I'm not sure that name makes sense: it isn't wrapping an existing type (which is what `as_raw` or `to_raw` would work great for), it's actually converting into a different type entirely.

Maybe it just shouldn't expose it at all, and we can defer the name decision to later (since `to_js` is mostly an implementation detail anyways).

fitzgen: > Maybe it just shouldn't expose it at all, and we can defer the name decision to later (since `to_js` is mostly an implementation detail anyways).

:+1:

crates/events/src/lib.rs
```diff
@@ -0,0 +1,521 @@
...
+    #[inline]
+    pub fn event_type(&self) -> &str {
+        &self.event_type
+    }
+
+    /// Returns the callback.
+    #[inline]
+    pub fn callback(&self) -> &Closure<FnMut(Event)> {
+        // This will never panic, because `callback` is always `Some`
+        self.callback.as_ref().unwrap_throw()
```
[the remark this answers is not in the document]
lnicola: That's what the timers do, I think.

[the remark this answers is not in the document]
fitzgen: forget takes `self` so once that is called, you wouldn't be able to use the listener again (it would be a use-after-move compile time error)

Pauan:
Just to clarify the status of this: I just need to put crate-level docs and add some docs to the guide and this'll be ready for review.

crates/events/src/lib.rs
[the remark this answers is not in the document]
Pauan: @thedodd It takes `mut self` not `&mut self` (easy to miss!)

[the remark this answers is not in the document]
Pauan: However, I'm glad you mentioned that, since it revealed a bug in the `Drop` implementation (which I just now fixed).

Pauan:
Okay, this is basically done, and ready for the final review.

I made some API changes since the previous review:

* Rather than using a `bool` for capture/bubble, I instead created an actual `EventListenerPhase` enum. I think this makes the API *much* clearer for users.

* I added in `Default` impls to `EventListenerOptions` and `EventListenerPhase`

* I added in some convenience methods to `EventListenerOptions` for common situations (capture, enabling `preventDefault`)

I added in unit tests, and docs for everything, but I wasn't sure what to put into the Gloo Guide... the other crates have an empty entry in the Guide, and wouldn't it just be repeating the crate-level docs...?

Pauan:
(By the way, my intention is to make the change from `FnMut(Event)` to `FnMut(&Event)` in a separate PR, since we have to wait for a wasm-bindgen release for that)

crates/events/tests/web.rs
```diff
@@ -0,0 +1,191 @@
...
+        body.click();
+        body.click();
+    })
+    .and_then(|results| {
+        is(results.len(), 1)?;
+        Ok(())
+    })
+}
+
+// TODO is it possible to somehow cleanup the closure after a timeout?
```
fitzgen: We could add `raw` getters for the underlying closure's `js_sys::Function`, and then use `web-sys` manually to remove the listener. This doesn't clean up the Rust closure, however. Probably that's alright (particularly since it isn't actually closing over anything, and therefore should just be a `fn` pointer).

crates/events/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+            callback: Some(callback),
+            phase,
+        }
+    }
+
+    /// Registers an event listener on an [`EventTarget`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.EventTarget.html).
+    ///
+    /// # Examples
+    ///
+    /// Registers a [`"click"`](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) event and casts it to the correct type
```
fitzgen:
```suggestion
    /// Registers a [`"click"`](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event) event and downcasts it to the correct `Event` subtype
```

crates/events/src/lib.rs
```diff
@@ -0,0 +1,541 @@
...
+    ///     // Stop the event from bubbling
+    ///     event.stop_propagation();
+    ///
+    ///     // ...
+    /// });
+    /// ```
+    ///
+    /// # Event type
+    ///
+    /// The event type can be either a `&'static str` like `"click"`, or it can be a `String`.
```
fitzgen:
```suggestion
    /// The event type can be either a `&'static str` like `"click"`, or it can be a dynamically constructed `String`.
```

crates/events/tests/web.rs
Pauan: > and then use web-sys manually to remove the listener.

Since it's attached to a `<div>` element which doesn't exist in the DOM, the event listeners themself will already be automatically cleaned up.

So the comment was just about the Rust Closure memory.

Pauan:
@fitzgen I debated whether to post this in the RFC thread, but since it's a (relatively) small change I decided to just go ahead and do it.

The `once` option in `EventListenerOptions` has bugged me for a while, and after a lot of thinking I came up with a solution:

1. I removed `once` from `EventListenerOptions`.
2. I added a new `once_with_options` method, which is just like `new_with_options` except it accepts `FnOnce` instead of `FnMut`.

I'm quite happy with this, now there is a nice symmetry:

* `new` -> `new_with_options`
* `once` -> `once_with_options`

And the `EventListenerOptions` can be used with either `new_with_options` or `once_with_options`.

I think this makes the API incredibly clear for users (it's much simpler now), and it also now allows for using `FnOnce` with options (that wasn't possible before).

(I also fixed the suggestions and added some more unit tests, so this is ready for the final final review.)

crates/events/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+        options: EventListenerOptions,
+        callback: F,
+    ) -> Self
+    where
+        S: Into<Cow<'static, str>>,
+        F: FnOnce(Event) + 'static,
+    {
+        let callback = Closure::once(callback);
+
+        Self::raw_new(
```
fitzgen: This doesn't actually ensure that the options have `once: true`, so the listener is not automatically unregistered and if the event is fired more than once it will panic. For example, the `"click"` listener in the above doc comment could fall into this trap.

I think we should either assert that `once: true` is set, force set it ourselves, or otherwise avoid this foot gun.

Pauan: @fitzgen Perhaps you are misunderstanding the code?

`once` is set in the `EventListenerOptions::to_js` method.

The `once_with_options` method calls `to_js(true)`, therefore `once` will always be `true`.

Remember, I *removed* `once` from `EventListenerOptions`, because now the "once-ness" is handled by `once_with_options` vs `new_with_options`

Also note that there is a unit test which does verify that `once_with_options` will only trigger the callback once.

crates/events/src/lib.rs
```diff
@@ -0,0 +1,594 @@
...
+    #[inline]
+    pub fn enable_prevent_default() -> Self {
+        Self {
+            passive: false,
+            ..Self::default()
+        }
+    }
+
+    #[inline]
+    fn to_js(&self, once: bool) -> AddEventListenerOptions {
```
fitzgen: Ah ok, I see now. Apologies my last review was submitted quickly before I had to run off on an errand.

If this was a public API, I'd advocate for replacing the bool with a more meaningful enum, but it isn't pub so I think the bool is fine.

fitzgen:
I think this is ready to merge now? Am I missing anything that makes it still [WIP] as the title implies?

Issue #37, opened 2019-03-20
XmlHttpRequest (mid-level only)


tomhoule:
## Summary

Provide "mid-level" bindings to the [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest) API. "mid-level" in this case would mean a callback-based API to cover the whole of XHR, mirroring `web_sys::XmlHttpRequest` with more idiomatic Rust types and ergonomic improvements where possible without compromising fidelity to the original.

## Motivation

The "blessed" browser API for HTTP going forward is `fetch`, but fetch does not provide all the features XMLHttpRequest does at the moment (one notable example being download/upload progress events).

This could also be used as a base for higher level bindings providing a more ergonomic interface from Rust, but I propose leaving it to a future discussion.

## Detailed Explanation

The bindings would be implemented inside a new `gloo-xhr` crate in a `callback` top-level module, to leave room for a potential future-based wrapper built on of it.

Two examples to motivate why we would benefit from "mid-level" bindings, compared to raw `web_sys`:

- Callbacks would be typed Rust closures.
- We wouldn't have many methods to set the body of a request, like in `web_sys`, but a `set_body` method generic over a (for example) `XhrRequestBody` trait implemented by all valid body types.

The documentation should mention that `fetch` is the API that is currently being pushed forward in browsers and will eventually supersede XmlHttpRequest.

## Drawbacks, Rationale, and Alternatives

Alternatively, these bindings could be grouped with the `fetch` bindings in a common `gloo-http` or `gloo-ajax` crate.

The implementation would aim to map as close as possible to the original web API; as such few design decisions should be necessary.

## Unresolved Questions

Providing nice, ideally typed errors could be challenging, as the error cases in networking APIs are many.

fitzgen:
Sounds good to me!

I like the idea of making a trait for request bodies.

PR #44, opened 2019-03-23
[WIP] Add mid-level bindings for XmlHttpRequest

tomhoule:
This is intended to implement the feature proposed in https://github.com/rustwasm/gloo/issues/37

TODO before this can be merged:

- Documentation.
- Realistic examples.
- Avoid leaking the event listeners (maybe use gloo-events?)
- More tests.
- More thoughtful error handling.

It would also be nice to figure out a way to spin up a test http server in the background for browser tests, but I haven't found a straightforward way to do it without running a separate command so far.


crates/xhr/src/lib.rs
```diff
@@ -0,0 +1,320 @@
...
+
+    /// See MDN docs.
+    #[derive(Debug, Clone)]
+    pub struct XmlHttpRequest {
+        xhr: web_sys::XmlHttpRequest,
+    }
+
+    impl XmlHttpRequest {
+        /// Initialize an XmlHttpRequest.
+        pub fn new() -> Self {
```
tomhoule: should we assume this is safe because all browsers that support webassembly implement XmlHttpRequest, or check whether the constructor is defined?

Pauan: If an API predates WebAssembly, then yes we can safely assume it exists.

In addition, [XMLHttpRequest goes all the way back to IE7](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#Browser_compatibility)! So we can safely assume it exists regardless.

Personally, I would just use `unwrap_throw`, but I don't have strong feelings about it.

fitzgen:
> Avoid leaking the event listeners (maybe use gloo-events?)

Yes, let's definitely build on top of `gloo_events` instead of using `onfoo` properties. This way, you can have more than one listener for each kind of event, and we get to exercise our own APIs a bit to make sure they're working smoothly.

tomhoule:
Just an update to say this PR isn't abandoned, I hope to find time to work on documenting and testing more soon. I am also waiting on gloo-events (I tried using it but gave up on panics in tests) to finish shaping the API.

tomhoule:
I think this is now ready for a bit more feedback. Here are a few doubts/questions I have regarding this implementation:

- is `impl AsRef<web_sys::EventTarget> for XmlHttpRequest` (meant for integration with gloo_events) a good idea? Would a regular method be better?
- The doctests crash because they can't run in non-wasm environments, should they be completely disabled?
- `send_no_body`: we could alternatively have `send()` and `send_with_body(body)`, or have `send` take an `Option<impl XhrBody>`. The second solution seems error-prone to me, but it's the only way to avoid having two methods. Is there a convention there?
- The test suite is very limited because I haven't figured out a way to spawn a real HTTP server in the background to test a lot more thoroughly. At least not without having a test script wrapping `wasm-pack test`.

tomhoule:
TODO: we can test abort events easily.

tomhoule:
It looks like the event handlers question is being discussed on #51, I'll read the conversation soon to adapt this.

fitzgen:
> is `impl AsRef<web_sys::EventTarget> for XmlHttpRequest` (meant for integration with gloo_events) a good idea? Would a regular method be better?

Seems good to me. Unlike somethign like `Deref` or `IntoIterator`, there can be more than one `AsRef` impl for a type (eg `AsRef<T>` and `AsRef<U>`) so it doesn't cut off other potential additions in the future.

> The doctests crash because they can't run in non-wasm environments, should they be completely disabled?

In these cases, I like to use <code>```no_run</code> blocks, which still verify that the example compiles, but does not attempt to run/test them.

> `send_no_body`: we could alternatively have `send()` and `send_with_body(body)`, or have `send` take an `Option<impl XhrBody>`. The second solution seems error-prone to me, but it's the only way to avoid having two methods. Is there a convention there?

I don't have strong opinions, but [this item in the API guidelines](https://rust-lang-nursery.github.io/api-guidelines/type-safety.html#arguments-convey-meaning-through-types-not-bool-or-option-c-custom-type) would suggest either two different methods or one method with a custom `enum`:

```rust
enum SendBody<B: XhrBody> {
    None,
    Body(B),
}

impl XmlHttpRequest {
    pub fn send<B: XhrBody>(&self, body: SendBody<B>) { ... }
}
```

We could also have the union of these methods:

```rust
impl XmlHttpRequest {
    pub fn send<B: XhrBody>(&self, body: SendBody<B>) { ... }
    pub fn send_with_body<B: XhrBody>(&self, body: B) { ... }
    pub fn send_without_body(&self) { ... }
}
```

Again, I don't really have strong opinions either way.

> The test suite is very limited because I haven't figured out a way to spawn a real HTTP server in the background to test a lot more thoroughly. At least not without having a test script wrapping `wasm-pack test`.

Another option is to load some JS snippet to monkey patch the API before the wasm and its JS glue imports it, and then mock up the behavior as needed:

```js
// gloo/crates/xhr/tests/mock.js

window.XMLHttpRequest.prototype.open = function(...) { ... };
window.XMLHttpRequest.prototype.send = function(...) { ... };

export function setNextRequestStatus(code) {
    // stash some state so that the next XHR we send will return the given HTTP status code..
}
```

```rust
// gloo/crates/xhr/tests/web.rs

#[wasm_bindgen(module = "/tests/mock.js")]
extern "C" {
    #[wasm_bindgen(js_name = setNextRequestStatus)]
    fn set_next_request_status(code: u32);
}
```

tomhoule:
Thanks for all the advice - there are improvements to make to this PR, I'll try to find the time to work on it asap, it's a bit complicated now as I am moving :)

crates/xhr/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+                .split("\r\n")
+                .map(|line| {
+                    let mut elems = line.split(": ");
+                    (
+                        elems.next().unwrap_throw().to_string(),
+                        elems.next().unwrap_throw().to_string(),
+                    )
+                })
+                .collect()
+        }
```
yoshuawuyts: Would it make sense for this to be an [`http::HeaderMap`](https://docs.rs/http/0.1.17/http/header/struct.HeaderMap.html) instead?

crates/xhr/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+            self.xhr
+                .send()
+                .expect_throw("Error sending request. Did you forget to call `open`?")
+        }
+
+        /// Actually send the request. In order to know the outcome, you have to attach
+        /// `load`, `abort` and `error` event listeners.
+        pub fn send_with_body<B: XhrBody>(&self, body: B) {
+            body.send(&self.xhr)
+                .expect_throw("Error sending request. Did you forget to call `open`?")
```
yoshuawuyts: I would've expected all the `send_*` methods to return `Result` instead of panicking. Could you expand on the reasoning here why panicking is preferred over `Result`?

yoshuawuyts:
I'm noticing there's quite a large API surface here, and I feel I have quite a bit of feedback I might want to share. Similar to the File API (https://github.com/rustwasm/gloo/pull/76), would it perhaps make sense to create consensus on the API in an RFC first, before moving to an implementation PR?

---

@tomhoule I understand proposing this might feel a bit frustrating, but I think if we do this well we might be able to land the implementation faster because we're splitting the design stage from the implementation allowing us to focus on one set of challenges at the time.

Also it might create a valuable resource to refer to when we inevitably want to tackle the Fetch / [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) APIs. Hope it's okay that I'm bringing this up!

crates/xhr/src/lib.rs
tomhoule: yep, this needs to be case insensitive, I realized writing tests for it

crates/xhr/src/lib.rs
tomhoule: I went the shortest route for error handling because there was no clear prior art/guideline there, but I agree that in principle a `Result` with a type error variant is ideal.

tomhoule:
sure - there's a lot to figure out, especially around error handling. I took up the PR thinking of getting this merged first and then polishing up, but we can discuss first (as you may have noticed I am very short on open source time at the moment though, sorry in advance if I can't get very involved). Let's close this. Should the discussion continue on https://github.com/rustwasm/gloo/issues/37?

Issue #23, opened 2019-03-13
Standardized documentation


David-OConnor:
## Summary

All Gloo modules should have narrative documentation, perhaps as part of a centralized guide

## Motivation

It's common in the Rust community to rely on API docs, and narrow-scoped examples. These are important, but this isn't always sufficient

## Detailed Explanation

With some exceptions, Rust libs, including most of the WASM frameworks tend to be documented through examples and API docs. While these are important, I think having a detailed, narrative guide is important too. For example, the Rust book. I think all Gloo components should have a maintained guide of this style. Could use the Rust Book template. Example: `Vulkano` has a very extensive API, a guide demonstrating some things, and a few examples, but is missing basic things, like how to use it for 3d items. Example: `Yew` is a popular stdweb-based framework, but has no narrative guide outside a brief readme, which until recently, was misaligned with the released API. Example: `Rocket` is a backend framework with a detailed guide.

Related: In addition to feature-specific examples, and long/complex ones, I think short ones demonstrating API coverage are useful... and missing in many Rust crates.

## Drawbacks, Rationale, and Alternatives

It requires more upfront effort, and maintenance work when APIs change.

fitzgen:
Thanks for bringing this up!

It isn't 100% clear to me whether you are suggesting

1. a guide to Gloo, the umbrella project, with sections for each of its crates plus whatever else fits, or
2. each crate within the toolkit has its own guide.

I am super down with (1) but I fear that (2) may be a bit all over the place, and make finding the right docs difficult.

Assuming we are on the same page with (1), would you be interested in submitting a PR to add a skeleton of an `mdbook` guide and a checkbox to `CONTRIBUTING.md` saying that each gloo crate should have a section(s) within the guide?

Aside: one pattern that has worked well for wasm-bindgen in terms of balancing maintainance costs and having lots of examples in a guide book is to use `mdbook`'s file inclusion features so that we can test example code and make sure it compiles.

David-OConnor:
1. Yes.

hamza1311:
I think the new website would be a great place for this documentation. I switched away from mdbook because there wouldn't have been enough content to call it a "book" and a website gave us more flexibly. I would love have the documentation worked up on the website as it's quite lacking at the moment.

anshulrgoyal:
@hamza1311 I am interested in adding documentation.

anshulrgoyal:
What is your vision for documentation on the website?

hamza1311:
[...]

PR #45, opened 2019-03-24
Guide skeleton

David-OConnor:
Reference https://github.com/rustwasm/gloo/issues/23 for a description. The intent is to provide a narrative, step-by-step description of how to use Gloo and each of its crates, in a way that complements API docs, examples, and module comments.

This is a preliminary skeleton: it includes an introduction, blank pages for the existing two crates. Once more modules are added, and Gloo's best-practices are solidified, we can expand it. The two existing crates are both narrow-scoped enough that narrative documentation isn't required, but more complicated crates will benefit from it.

The introduction is deliberately brief, in order not to duplicate the readme.

### DIFF
--- a/guide/book.toml
+++ b/guide/book.toml
@@ -1,5 +1,5 @@
 [book]
-authors = ["David O'Connor"]
+authors = ["rustwasm"]
 multilingual = false
 src = "src"
 title = "Gloo Guide"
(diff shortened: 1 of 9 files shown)


guide/book.toml
```diff
@@ -0,0 +1,5 @@
+[book]
+authors = ["rustwasm"]
```
fitzgen:
```suggestion
authors = ["The Rust and WebAssembly Working Group"]
```

Issue #50, opened 2019-03-25
Mid-Level File API Proposal


rylev:
Addresses #47 in part by proposing a mid-level API. The possibility for a higher level API is left open. 

# File API

The mid-level file API aims to implement File I/O on top of the raw JS apis found here: https://w3c.github.io/FileAPI/ which includes listing files, creating files, and reading files.

## General Differences 

In general this API will mostly differ from the `web_sys` interface in the following ways:

### No js_sys Types

js_sys tends to be fairly "stringly" typed, so where we have known enums, the crate will provide bespoke types.  

This is also true for errors which are usually exposed as JsValues. In this crate they will be exposed as well-formed Rust types.

### Callbacks 

All callbacks will be of the appropriate closure type (e.g., FnOnce).

### FileList

JS provides a basic collection type over files (usually from an <input> tag). In general `FileList` will be an opaque struct that offers a rich interface to the collection through an `Iterator` implementation. 

###  Blob

`Blob` is a "file like" interface. The crate would expose this as a Rust interface from which the specific types `File` and `RawData` inherit. These types would be roughly 1 to 1 translations of the `File` and `Blob` objects just with less "stringly" typed interfaces. For example, `RawData` would accept a known list of mime-types 

### FileReader

The FileReader API is a bit wonky and can easily be used in the incorrect way. For instance, if the user calls `readAsArrayBuffer` or other "read" APIs before the data is loaded, the operation won't return data. The crate will expose an API that more properly models the "read" state machine. 

This will be done through a futures API which takes the user through the state machine:

```rust
let file_reader = FileReader::new();
file_reader.read_as_text(&blob).map(move |contents| {
  element.set_inner_html(contents);
})
```

The various "read" methods will take ownership of `self` so that users can't try to read multiple files at once. These methods return an `InProgressRead` struct that is a future and also contains an `abort` method for aborting the upload.

The events "error", "load", and "loadend" are all representable through the future and thus aren't exposed on the file reader. The events "progress", "loadstart", and "abort" are still exposed as methods that take the appropriate Rust closure. 

### FileReaderSync

This API would mirror the future based API just in a different namespace. 

## Drawbacks, Rationale, and Alternatives

FileReader exposes a very different API from its JS counterpart. We could do a more 1 to 1 translation of the API and leave the improved API for a higher-level crate. This would make this crate only a very marginal improvement on what is already exposed in `js_sys`.

## Unresolved Questions

[...]

PR #51, opened 2019-03-26
Files API Implementation

rylev:
Aims to implement #50 

This is still work-in-progress and the discussion on the API proposal is not complete, but I thought the conversation could be enriched through some code.


crates/file/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+  pub fn new() -> FileReader {
+    FileReader {
+      inner: web_sys::FileReader::new().unwrap(),
+    }
+  }
+
+  pub fn read_as_string(
+    self,
+    blob: &(impl Blob + RawBlob),
+  ) -> impl futures::Future<Item = String, Error = ()> {
```
rylev: I think we'll want to create our own Future type that cleans up the closure when it is dropped. As the code stands right now, if the future is dropped before the reader finishes reading the blob, we'll leak memory.

crates/file/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+  inner: web_sys::File,
+}
+
+impl File {
+  fn from_raw(inner: web_sys::File) -> File {
+    File { inner }
+  }
+}
+
+pub enum MimeType {
```
rylev: We'll reuse https://crates.io/crates/mime instead of having our own types.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub fn new() -> FileReader {
+        FileReader {
+            inner: web_sys::FileReader::new().unwrap(),
+        }
+    }
+
+    pub fn read_as_string(
+        self,
+        blob: &impl Blob,
+    ) -> impl futures::Future<Item = String, Error = ()> {
```
fitzgen: Rather than using `impl Future` here, we should define a proper `Future` type. While using `-> impl Trait` is often convenient for the author, it is less convenient for users (since they can't use the type as a field of a struct, often aren't `Debug` or `Clone`).

Also, the struct is the thing that should contain the `Closure`, to get the lifetimes right.

```rust
#[derive(Debug)] // will actually have to be a manual impl until wasm-bindgen#1387 is fixed
pub struct ReadAsString {
    receiver: oneshot::Receiver<String>,
    closure: Closure<FnMut(...)>,
}

impl Future for ReadAsString { ... }
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,37 @@
...
+        let (tx, rx) = futures::sync::oneshot::channel();
+        let reader = Rc::new(self.inner);
+        let cloned_reader = reader.clone();
```
fitzgen: Cloning the `inner` should be ~about as fast as `Rc` cloning, but won't force us to include all the `Rc` code in the wasm binary.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,37 @@
...
+        let cb = wasm_bindgen::closure::Closure::once_into_js(move || {
+            let _ = cloned_reader.result().map(|r| {
+                let _ = tx.send(r.as_string().unwrap());
```
fitzgen: Nitpick: `unwrap_throw`

crates/file/src/blob.rs
```diff
@@ -0,0 +1,44 @@
+pub use crate::file::File;
+
+pub trait Blob: raw::Blob {
+    fn size(&self) -> usize {
+        self.raw().size() as usize
+    }
+
+    fn mime_type(&self) -> Option<mime::Mime> {
+        Some(self.raw().type_().parse().ok()?)
```
fitzgen: Have you looked at the difference in code size between returning the raw string vs using the `mime` crate to parse? If it is sizable, we probably want to different methods, and maybe a feature for `mime`

crates/file/src/blob.rs
```diff
@@ -0,0 +1,44 @@
+pub use crate::file::File;
+
+pub trait Blob: raw::Blob {
```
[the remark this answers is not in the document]
rylev: I want the trait to be sealed so that others don't implement it. Is there a better way to do that than this?

crates/file/src/file_list.rs
```diff
@@ -0,0 +1,57 @@
...
+            current: 0,
+        }
+    }
+
+    pub fn into_vec(self) -> Vec<File> {
+        self.iter().collect()
+    }
+}
+
+pub struct FileListIter<'a> {
```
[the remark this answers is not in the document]
rylev: Totally. Will work on those next.

crates/file/src/file_reader.rs
rylev: Agree 100%.

crates/file/src/blob.rs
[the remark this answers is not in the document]
fitzgen: /me shrugs

I don't think it is that big of a deal: the expectation of bad outcomes for users implementing this is much lower than the expectation of bad outcomes for users getting confused by a super trait that they can't read the docs for, IMO.

rylev:
So we're in a good minimal place. The following things are left to implement and/or discuss:
* The other FileReader methods: `.readAsDataURL()`, `readAsBinaryString()`, and `readAsArrayBuffer()` 
  * Should dataUrl and binaryString be strings? Should gloo have a higher level ArrayBuffer object or should we copy the file into the wasm heap and return a Vec<u8> to the user?  What about other integer sizes?
* Add cancelation support to `ReadAsString` future. 
* Fine grained support of different events on the `FileReader` (e.g., `loadstart` and `progress`)
  * Should these only be exposed as a high level stream API. What about callbacks?
* Add file metadata (e.g., name) getters 
* Add a slice interface on BlobLike things

samcday:
> Should gloo have a higher level ArrayBuffer object or should we copy the file into the wasm heap and return a Vec to the user?

It should be sufficient to just provide the `Uint8Array`, right? The API web_sys provides for it is already rich enough. If the caller wants to copy that into Rust land, it's a mere [`Uint8Array::copy_to`](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Uint8Array.html#method.copy_to) away.

fitzgen:
`readAsBinaryString` seems deprecated, its functionality is subsumed by `readAsArrayBuffer`, and it only exists for backwards compatibility -- I say we ignore it.

`readAsArrayBuffer` -- I don't think we want to *force* a copy into wasm memory. If for whatever reason the user is just going to pass it back to some other JS method, then that copy would be a big waste. The lowest common denominator API is returning the `js_sys::ArrayBuffer` directly. On top of that, we could provide convenience methods to do the typed-array `copy_to` dance for different integer sizes and signednesses. Would require a little care so that we allowed folks to bring their own buffers rather than forcing new `Vec` allocations.

>     * Add cancelation support to `ReadAsString` future.

I haven't looked at the updated code yet, but this *should* be automatic on drop.

> Fine grained support of different events on the `FileReader` (e.g., `loadstart` and `progress`)
> 
>     * Should these only be exposed as a high level stream API. What about callbacks?

I think we can:

* provide a constructor or builder that takes optional callbacks for these events (using `set_onprogress` rather than `gloo_events`)

* allow folks to add arbitrary listeners to the underlying event target

* in a backwards-compatible way, add methods for getting the progress events as a Stream (this can be delayed to a new PR; we probably want to add a generic event listener -> stream conversion infrastructure to build this on top of)

crates/file/src/blob.rs
```diff
@@ -0,0 +1,60 @@
+use wasm_bindgen::UnwrapThrowExt;
+
+pub trait BlobLike {
+    fn size(&self) -> usize {
+        self.as_raw().size() as usize
+    }
+
+    #[cfg(feature = "mime")]
+    fn mime_type(&self) -> Option<mime::Mime> {
+        Some(self.as_raw().type_().parse().ok()?)
```
fitzgen: I think we should return the `Result` that `parse()` returns, rather than converting it to an `Option`

Pauan:
> provide a constructor or builder that takes optional callbacks for these events (using set_onprogress rather than gloo_events)


Wouldn't they have to manually handle all the Closure related stuff? Why not just use `gloo-events`?

fitzgen:
> Wouldn't they have to manually handle all the Closure related stuff? Why not just use `gloo-events`?

I was imagining that there would be `onprogress: Option<Closure<FnMut()>>` inside `FileReader` (rather than a `Vec<EventListener>`)

But yeah I'm not super happy about it, and all the alternatives seem to have different downsides.

Alternatives for progress would be:

* Only exposing the event target for users to manually add callbacks. Now they have to manage the `EventListener` separately from the `FileReader`. Maybe this is OK.
* Not exposing callbacks at all, and just the streams of events. Goes against our layering principles, but maybe is worth it in certain cases. Does mean that folks *must* use streams to get progress events.
* Storing a `Vec<EventListener>` inside `FileReader`. These would probably be add-only since (as I believe you mentioned in other threads) there is no good way to remove a specific one without introducing a whole new RAII type or Rc or something.

Pauan:
> I was imagining that there would be `onprogress: Option<Closure<FnMut()>>` inside FileReader (rather than a `Vec<EventListener>`)

Why can't it be `onprogress: EventListener` or `onprogress: Option<EventListener>`? There's no need for `Vec` if you know ahead of time what events you want.

The reason I mentioned `Vec<EventListener>` in the other thread was because that API actually *wanted* the ability to dynamically add/remove arbitrary events (not just a hardcoded list).

fitzgen:
Oh yeah totally. That's what we should do. `<brain-fart>`

rylev:
I've made good progress on the todos above. I need to wait for the Events PR to be merged to finish this out though.

@fitzgen I want to provide a `slice` interface for BlobLike things. The JavaScript interface uses signed integers to indicate start and end indices (e.g., -10 for the end value is 10 away from the end). This convention isn't too common in Rust. Do you think we should stick with the JavaScript convention or do something more straightforward (signed integers that just specify beginning and end indices?

fitzgen:
We should move towards Rust idioms, not JavaScript idioms. Therefore, I do not think we should do the signed-index-with-negative-indexing-from-the-end thing.

rylev:
@fitzgen @Pauan I'm going to close this for now until we close on the API proposal. Once that's decided I can open this up again.

PR #54, opened 2019-03-27
Run headless browser tests in CI

fitzgen:
@rylev or @alexcrichton want to take a look at this?

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -18,20 +18,20 @@ jobs:
     steps:
       - template: .ci/install-rust.yml
       - template: .ci/install-wasm-pack.yml
-      - displayName: "Run `wasm-pack test`"
-        script: |
+      - script: |
           for x in $(ls crates); do
             wasm-pack test --headless --firefox crates/$x -- --all-features || exit 1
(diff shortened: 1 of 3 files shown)


fitzgen:
Hmm... why isn't CI getting triggered?

alexcrichton:
Oh [looks like it did](https://dev.azure.com/rustwasm/gloo/_build/results?buildId=279&view=results) but maybe it's a bug in azure pipelines that the status there wasn't reported back here?

.azure-pipelines.yml
```diff
@@ -10,6 +11,27 @@ jobs:
           rust_version: stable
       - bash: |
           export RUST_BACKTRACE=1
-          cargo check --all || exit
           cargo test --all || exit
-        displayName: Run cargo check and test 
+        displayName: Run cargo check and test
```
alexcrichton: Was there maybe some whitespace changes here?

fitzgen: Just removing a trailing space -- I think the issue is `displayName` coming before `script`? Trying to swap the order...

fitzgen:
Phew! Ok 5 crappy "fixing CI (again)" style commits and now our headless browser tests are running in CI!

PR #56, opened 2019-03-28
Fix timers in web workers

samcday:
@fitzgen / @OddCoincidence 

`gloo-timers` doesn't work in web workers at all right now, because `web_sys::window()` wants to cast the global namespace into a `Window` object, which doesn't exist in worker contexts. So I just dropped us back down to raw `#[wasm_bindgen]` methods.

This might be something better fixed in wasm-bindgen later though. There does exist this thing: [WindowOrWorkerGlobalScope](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope) which is a nice view over what global methods are available in both Window and Worker contexts. However as stated in that documentation:

> Note: WindowOrWorkerGlobalScope is a mixin and not an interface; you can't actually create an object of type WindowOrWorkerGlobalScope.

So I guess this thing is probably not mentioned anywhere in the IDLs that are being used to build the `web-sys` crate.

~Also, the `Interval` stuff wasn't actually working properly at all, it was consuming the callback from the `Option` which meant only the first interval callback worked and all subsequent ones failed.~ See #57

### DIFF
--- a/crates/timers/src/lib.rs
+++ b/crates/timers/src/lib.rs
@@ -67,25 +67,9 @@ TODO
 #[cfg(feature = "futures")]
 extern crate futures_rs as futures;
 
-use js_sys::Function;
-use wasm_bindgen::prelude::*;
-
-#[wasm_bindgen]
-extern "C" {
-    #[wasm_bindgen(js_name = "setTimeout")]
-    pub fn set_timeout(handler: &Function, timeout: i32) -> i32;
-
-    #[wasm_bindgen(js_name = "clearTimeout")]
-    pub fn clear_timeout(token: i32);
-
-    #[wasm_bindgen(js_name = "setInterval")]
-    pub fn set_interval(handler: &Function, timeout: i32) -> i32;
-
-    #[wasm_bindgen(js_name = "clearInterval")]
-    pub fn clear_interval(token: i32);
-}
-
 pub mod callback;
 
 #[cfg(feature = "futures")]
 pub mod future;
+
+mod sys;
--- a/crates/timers/src/lib.rs
+++ b/crates/timers/src/lib.rs
@@ -67,10 +67,22 @@ TODO
 #[cfg(feature = "futures")]
 extern crate futures_rs as futures;
 
+use js_sys::Function;
 use wasm_bindgen::prelude::*;
 
-fn window() -> web_sys::Window {
-    web_sys::window().unwrap_throw()
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_name = "setTimeout")]
+    pub fn set_timeout(handler: &Function, timeout: i32) -> i32;
+
+    #[wasm_bindgen(js_name = "clearTimeout")]
+    pub fn clear_timeout(token: i32);
+
+    #[wasm_bindgen(js_name = "setInterval")]
+    pub fn set_interval(handler: &Function, timeout: i32) -> i32;
+
+    #[wasm_bindgen(js_name = "clearInterval")]
+    pub fn clear_interval(token: i32);
 }
 
 pub mod callback;
--- a/crates/timers/src/callback.rs
+++ b/crates/timers/src/callback.rs
@@ -1,6 +1,6 @@
 //! Callback-style timer APIs.
 
-use super::{clear_interval, clear_timeout, set_interval, set_timeout};
+use super::sys::*;
 use std::fmt;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
--- a/crates/timers/src/callback.rs
+++ b/crates/timers/src/callback.rs
@@ -1,6 +1,6 @@
 //! Callback-style timer APIs.
 
-use super::window;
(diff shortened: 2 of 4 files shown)


OddCoincidence:
> Also, the Interval stuff wasn't actually working properly at all, it was consuming the callback from the Option which meant only the first interval callback worked and all subsequent ones failed.

Sorry about this! That's a copy-paste error from timeout (from before it was using `Closure::once`). I'm surprised / concerned that a test didn't catch this.

samcday:
> Sorry about this! That's a copy-paste error from timeout (from before it was using Closure::once).

Apology accepted. Don't let it *ever* happen again. Now go to your room and put away your clean washing. I want to see the socks FOLDED this time.

> I'm surprised / concerned that a test didn't catch this.

Yeah interesting you should say that. I haven't looked at the test suite much but I have some concerns with it too. Namely when I made these changes I ran the test suite and everything passed. But then for the lolz I changed the `js_name = setTimeout` to `js_name = spaghetti` and ... everything still passed. You might wanna look into that more :)

> Maybe these two issues should be split into separate PRs though since they aren't related?

Sure - I was just being lazy ;) I'll split the interval fix out now.

samcday:
> But then for the lolz I changed the js_name = setTimeout to js_name = spaghetti and ... everything still passed. You might wanna look into that more :)

Actually no ... ignore me, I was only running `cargo test` and not the web tests. I'm actually not sure how to run the integration tests, this is what I get:

```
sam@SamsPC:~/src/gloo/crates/timers$ wasm-pack test --firefox --headless
  
  [1/6] Checking `rustc` version...
  [2/6] Adding WASM target...
  info: component 'rust-std' for target 'wasm32-unknown-unknown' is up to date
  [3/6] Compiling tests to WASM...
     Compiling gloo-timers v0.1.0 (/home/sam/src/gloo/crates/timers)
      Finished dev [unoptimized + debuginfo] target(s) in 0.19s
  [4/6] Installing wasm-bindgen...
  [5/6] Getting geckodriver...
  [6/6] Running tests in Firefox...
      Finished dev [unoptimized + debuginfo] target(s) in 0.07s
       Running /home/sam/src/gloo/target/wasm32-unknown-unknown/debug/deps/gloo_timers-723cf89fd2565a10.wasm
  no tests to run!
       Running /home/sam/src/gloo/target/wasm32-unknown-unknown/debug/deps/web-fbd694454b48e118.wasm
no tests to run!                                    
| 
no tests to run!
no tests to run! 
```

OddCoincidence:
Ah, I think this is because the integration tests are gated on the recently-added `futures` feature. I think `wasm-pack  test --chrome --headless -- --features futures` ought to do the trick (everything after the `--` should be passed to `cargo test`).

samcday:
`wasm-pack test --chrome --headless -- --features futures` didn't work either:

```
sam@SamsPC:~/src/gloo/crates/timers$ wasm-pack test --firefox --headless -- --features futures
error: Found argument 'futures' which wasn't expected, or isn't valid in this context

USAGE:
    wasm-pack test --firefox --headless

For more information try --help
```

OddCoincidence:
What's your version of `wasm-pack`? Passing flags to `cargo test` was added in 0.7

samcday:
Ah good catch, I'm on 0.6 still.

Pauan:
> gloo-timers doesn't work in web workers at all right now, because web_sys::window() wants to cast the global namespace into a Window object, which doesn't exist in worker contexts.

Relevant: https://github.com/rustwasm/wasm-bindgen/issues/1046

crates/timers/src/lib.rs
```diff
@@ -67,10 +67,22 @@ TODO
...
 
+use js_sys::Function;
 use wasm_bindgen::prelude::*;
 
-fn window() -> web_sys::Window {
-    web_sys::window().unwrap_throw()
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(js_name = "setTimeout")]
+    pub fn set_timeout(handler: &Function, timeout: i32) -> i32;
```
fitzgen: Rather than taking `&js_sys::Function`, if we're going to write the bindings by-hand, we should make them take `&Closure<FnMut()>`

Also, let's put these raw bindings into a `sys` module.

fitzgen:
Regarding the headless browser tests: as of #54 they are running in CI, so we should be better protected going forward.

crates/timers/src/lib.rs
samcday: I moved the bindings to a `sys` module, but I'm not sure about changing the signature to accept a Closure. The problem is the `callback` module is using `js_sys::Promise` and passing the `resolve: &Function` parameter straight into the set_timeout. It was already doing that before, since the main web_sys api for [`set_timeout_with_callback_and_arguments`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.WorkerGlobalScope.html#method.set_timeout_with_callback_and_timeout_and_arguments) accepts a `Function`.

fitzgen: Ah, this is a good point; I forgot that it was using promises like that.

PR #57, opened 2019-03-28
gloo-timers: don't consume callback in Interval closure

samcday:
Otherwise Interval only works on first callback and fails after that

### DIFF
--- a/crates/timers/src/callback.rs
+++ b/crates/timers/src/callback.rs
@@ -152,11 +152,7 @@ impl Interval {
     where
         F: 'static + FnMut(),
     {
-        let mut callback = Some(callback);
-        let closure = Closure::wrap(Box::new(move || {
-            let callback = callback.as_mut().take().unwrap_throw();
-            callback();
-        }) as Box<FnMut()>);
+        let closure = Closure::wrap(Box::new(callback) as Box<FnMut()>);
 
         let id = window()
             .set_interval_with_callback_and_timeout_and_arguments_0(
--- a/crates/timers/src/callback.rs
+++ b/crates/timers/src/callback.rs
@@ -154,7 +154,7 @@ impl Interval {
     {
         let mut callback = Some(callback);
         let closure = Closure::wrap(Box::new(move || {
-            let mut callback = callback.take().unwrap_throw();
+            let callback = callback.as_mut().take().unwrap_throw();
             callback();
         }) as Box<FnMut()>);
 
--- a/crates/timers/tests/web.rs
+++ b/crates/timers/tests/web.rs
@@ -91,53 +91,57 @@ fn timeout_future_cancel() -> impl Future<Item = (), Error = wasm_bindgen::JsVal
 #[wasm_bindgen_test(async)]
 fn interval() -> impl Future<Item = (), Error = wasm_bindgen::JsValue> {
     let (mut sender, receiver) = mpsc::channel(1);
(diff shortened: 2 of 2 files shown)


crates/timers/src/callback.rs
```diff
@@ -154,7 +154,7 @@ impl Interval {
     {
         let mut callback = Some(callback);
         let closure = Closure::wrap(Box::new(move || {
-            let mut callback = callback.take().unwrap_throw();
+            let callback = callback.as_mut().take().unwrap_throw();
```
OddCoincidence: I think this could be simplified to:
```
let closure = Closure::wrap(callback);`
```

OddCoincidence:
Could we also add a test that triggers this bug?

crates/timers/src/callback.rs
Pauan: Because of `take()`, I'm pretty sure this will still break.

So I agree with @OddCoincidence, that this should just completely remove the `Some` wrapper and use `Closure::wrap(Box::new(callback) as Box<FnMut()>)` instead

samcday: Yeah I already updated my local copy to `Closure::wrap(Box::new(callback) as Box<FnMut()>)`, I'm just grappling with the integration tests at the moment and should have the changes pushed up soon.

samcday: Also the current version with `.take()` does work and is re-entrant. It works because `as_mut` is creating a new Option that is a mutable reference. So `take()` on a `.as_mut()` is not consuming the original Option.

samcday:
@OddCoincidence yep working on tests now. As it happens integration tests are not running in CI, which is why it's reporting that these changes are okay here, when in fact fixing the bug actually causes the `interval_cancel` test to fail.

samcday:
@OddCoincidence alright I wrangled the tests. Hope you don't mind but I did a little bit of re-jiggery once I understood how they worked to simplify them a little.

The `interval` test now ensures that 5 ticks of the interval pass. `interval_stream` was already doing that so now they're consistent. I removed the `Cell` stuff in favour of just `collect()`ing the Stream and counting how many results we got.

The `interval_cancel` test started failing once I fixed the original bug. I reworked it a little to avoid the dance it was doing in favour of just combining a Timeout stream with the `Interval` stream we're examining.

crates/timers/tests/web.rs
```diff
@@ -91,53 +91,57 @@ fn timeout_future_cancel() -> impl Future<Item = (), Error = wasm_bindgen::JsVal
 #[wasm_bindgen_test(async)]
 fn interval() -> impl Future<Item = (), Error = wasm_bindgen::JsValue> {
     let (mut sender, receiver) = mpsc::channel(1);
-    Interval::new(1, move || sender.try_send(()).unwrap()).forget();
+    Interval::new(1, move || {
+        if !sender.is_closed() {
+            sender.try_send(()).unwrap()
+        }
+    })
+    .forget();
```
fitzgen: We shouldn't forget this -- that will cause the `setInterval` call to keep running even after the test is done.

Instead, we should do `let interval = Interval::new(...)` and then in the `and_then` closure, we should `drop(interval)` so that it gets cleared once the test is finished.

samcday: Ah yes, good call. Done!

samcday:
Sweet, feedback addressed. Looks like CI is running headless tests now too, nice!

PR #61, opened 2019-03-30
WIP: notifications crate

Aloso:
Displaying notifications on the web.

### This is a work in progress and hasn't been tested. Feedback is very welcome.

This API comes in two flavors: A callback style and `Future`s API.

Before a notification can be displayed, the user of the browser has to give his permission.

## 1. Callback style

```rust
use gloo_notifications::Notification;

Notification::request_permission_map(|mut builder| {
    builder.title("Notification title").show();
});

Notification::request_permission_map_or(|mut builder| {
    builder.title("Notification title")
        .body("Notification body")
        .show();
}, |_| {
    // in case the permission is denied
});

// short form, if you only need a title
Notification::request_permission_with_title("Notification title");
```

## 2. `Future` API:

```rust
use gloo_notifications::Notification;

Notification::request_permission()
    .map(|mut builder| {
        builder.title("Notification title").show();
    })
    .map_err(|_| {
        // in case the permission is denied
    })
```

## Adding event listeners

I have only implemented `onclick` so far, since I don't know how event listeners should be implemented.

```rust
use gloo_notifications::Notification;

Notification::request_permission_map(|mut builder| {
    let notification = builder
        .title("Notification title")
        .show();
    notification.onclick(|_| { ... });
})
```
## Macro
```rust
use gloo_notifications::{Notification, notification};

// requests permission, then displays the notification
// and adds a "click" event listener
notification! {
    title: "Hello World",
    body: "Foo",
    icon: "/assets/notification.png";
    onclick: |_| {}
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,9 +10,10 @@ version = "0.1.0"
 [dependencies]
 gloo-timers = { version = "0.1.0", path = "crates/timers" }
 gloo-console-timer = { version = "0.1.0", path = "crates/console-timer" }
+gloo-notifications = { version = "0.1.0", path = "crates/notifications" }
 
 [features]
 default = []
-futures = ["gloo-timers/futures"]
+futures = ["gloo-timers/futures", "gloo-notifications/futures"]
 
 [workspace]
--- a/crates/notifications/Cargo.toml
+++ b/crates/notifications/Cargo.toml
@@ -1,0 +1,36 @@
+[package]
+name = "gloo-notifications"
+version = "0.1.0"
+authors = ["Rust and WebAssembly Working Group"]
+edition = "2018"
+
+[dependencies]
+wasm-bindgen = "0.2.40"
+js-sys = "0.3.17"
+
+[dependencies.futures_rs]
+package = "futures"
+version = "0.1.25"
+optional = true
+
+[dependencies.wasm-bindgen-futures]
+version = "0.3.17"
+optional = true
+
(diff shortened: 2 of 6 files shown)


DesmondWillowbrook:
Looking at the usage, it looks like this API is not mirroring [the JS one](https://developer.mozilla.org/en-US/docs/Web/API/notification) and rather adding a lot of its own API?

hamza1311:
> Looking at the usage, it looks like this API is not mirroring [the JS one](https://developer.mozilla.org/en-US/docs/Web/API/notification) and rather adding a lot of its own API?

A mirrored API already exists (see `web_sys`). `gloo` is supposed to provide an API that is idiomatic Rust, not an API that that mirrors the JS one

DesmondWillowbrook:
I agree that the API is intended to be Rust-like, but my impression of the current aim (for the low-level libraries in Gloo) was to remove any bindgen stuff and use only Rust native types in the APIs (and remove extraneous `Result`s from `web_sys`, wrap JS errors in proper error enums, etc.), _then_ advance to middle-to-high level APIs which are "opinionated", i.e. make stuff ergonomic.

Specifically, looking at the builder pattern which is implemented here or the notification macro makes it feel like a higher level API.

These additions certainly look helpful, just that I don't think that the first implementation for notifications into `gloo` should be even that opinionated. Maybe once the JS API surface has been Rustified, then we can look at building it further.

Issue #21, opened 2019-03-13
WebExtensions crate


Pauan:
## Summary

A crate which contains idiomatic high-level Rust APIs for the [WebExtensions standard](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions), which is used to create browser extensions for Firefox, Chrome, and Edge.

## Motivation

I personally have made multiple Chrome extensions, so this is something I need and have a lot of experience with.

Being able to create *fast* and *safe* and *maintainable* browser extensions in Rust is a big deal. In my experience JavaScript just does not scale well with big extensions.

But the APIs are... well, rather terrible. So in addition to porting the APIs to Rust, we should clean them up and make them actually reasonable.

I've done this sort of work before in JavaScript, so I know a lot of the weird quirks and gotchas of the APIs.

OddCoincidence:
I would like to work on this, @Pauan would you be willing to mentor?

I assume the first step is to add support for `WebExtensions` to `wasm-bindgen`, perhaps a new IDL-generated `webextensions-sys` crate?

dlight:
There is [web-extensions](https://github.com/web-extensions-rs/web-extensions) which uses [web-extensions-sys](https://github.com/web-extensions-rs/web-extensions-sys) now

PR #62, opened 2019-03-31
[WIP] WebExtensions crate

OddCoincidence:
Sharing what I have right now for #21. Any feedback will be very much appreciated! Thanks to @Pauan for their help so far.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,7 @@
 **/*.rs.bk
 Cargo.lock
 
+node_modules
+pkg
+
 guide/book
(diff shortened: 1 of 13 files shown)


fitzgen:
Thanks for creating a PR @OddCoincidence.

However, I'd like to see more [design discussion and an API proposal](https://github.com/rustwasm/gloo/blob/d71b861192d6049a3ddc7928dae1cb69cbf5b888/.github/ISSUE_TEMPLATE/propose_design.md) before we rush to PRs. That is, I'd like ideally PRs to be unsurprising and where we don't ahve any design kinks to work out, and we do all the drafting and design discussion in the issue thread beforehand.

OddCoincidence:
Ah, okay, that's fair. I'll just close this then.

PR #63, opened 2019-04-01
Add a bit about the API's skeleton to the API proposal template

fitzgen:


fitzgen:
This is the thing I often want to see most when looking at API proposals, so we should nudge people towards providing it :)

PR #65, opened 2019-04-01
Expand and formalize contribution workflow a bit

fitzgen:
This lays out / formalizes the workflow I've had in my head for designing APIs, implementing them, and contributing to Gloo. Happy to use this as a starting point and modify it however. I figured having the graph visualization would be good for getting the idea across quickly.

What do you think?

cc @Pauan @rylev @yoshuawuyts @David-OConnor @samcday


fitzgen:
Direct link to visualization: https://raw.githubusercontent.com/rustwasm/gloo/e87ca651b6e5c97574935b21940c01beb0221359/new-design-workflow.png

CONTRIBUTING.md
```diff
@@ -129,32 +160,80 @@ Here is a checklist that all Gloo utility crates should fulfill:
       escape hatch for dropping down to raw `web_sys` bindings when an API isn't
       fully supported by the crate yet.
 
-* [ ] There is a loose hierarchy with "mid-level" APIs (which are essentially thin wrappers over the low-level APIs), and "high-level" APIs (which make more substantial changes).
+      Similar for `from_fraw` constructors and `into_raw` conversion methods
```
Pauan: Nit: `from_raw`

CONTRIBUTING.md
```diff
@@ -129,32 +160,80 @@ Here is a checklist that all Gloo utility crates should fulfill:
...
+
+The implementation should generally be unsurprising, since we should have
+already worked out most of the kinks during the earlier design discussions. If
+there are significant new issues or concerns uncovered during implementation,
+then these should be brought up in the design proposal discussion thread again,
+and the evolved design reaffirmed with two team members signing off once
+again.
+
+If there are no new concerns uncovered, then the implementation just needs to be
+checked over by at least one team member. They provide code review and feedback
```
Pauan: I don't think it will be a problem, but I'd like to note that this *does* allow for one team member to approve the pull request while another disapproves.

And this also allows for the team member who approves the pull request to be *different* from the team members who OKed the design.

new-design-workflow.dot
```diff
@@ -0,0 +1,94 @@
+digraph {
```
Pauan: Bit off topic, but how did you create this?

CONTRIBUTING.md
samcday: > but I'd like to note that this does allow for one team member to approve the pull request while another disapproves.

I didn't get that impression. You only get to the "Merge" state if all pending feedback has been addressed. So if a team member says "please don't merge this until a test for <foo behaviour>" has been written, that needs to be done before a different team member could approve and merge.

I guess it doesn't handle the (in my mind unlikely) scenario in which a team member is generally concerned about the PR overall. Like if they felt the implementation was just going in completely the wrong direction or something. But even in those cases, I would make the case that nobody should be saying "I *disapprove* of these changes!" without some concrete reason why. And those concrete reasons fall into the Changes Requested state.

samcday:
This is great, thanks a lot for doing this.

CONTRIBUTING.md
Pauan: I was thinking more about the situation with team members being away for a while (or not having enough time to even look at the design).

Like I said though, I don't think it'll be a problem in practice. Just something to keep in mind.

new-design-workflow.dot
fitzgen: [Graphviz](https://graphviz.org/)'s `dot` CLI tool. It's awesome. I'll add a comment that describes how to re-build the image from source.

CONTRIBUTING.md
fitzgen: Fixed.

CONTRIBUTING.md
fitzgen: I added a note that if any team member has concerns with an implementation, then that must be resolved before merging. I suspect that most of this stuff will already be worked out during the design proposal phase, though.

fitzgen:
> I'm kind of curious about a "recommendations section" for the last mile improvements

We have some of these now (debug impls and docs). I think it is worth spinning the others off into another issue for discussion (since this PR is mostly about the workflow / rules of what constitutes acceptance of a design)

CONTRIBUTING.md
```diff
@@ -92,34 +96,61 @@ To (re)format the Gloo source code, run:
...
 
-## Gloo Crate Checklist
+## Workflow
 
-Here is a checklist that all Gloo utility crates should fulfill:
+Designing APIs for Gloo, its utility crates, and interfaces between them takes a
+lot of care. The design space is large, and there is a lot of prior art to
+consider. When coming to consensus on a design, we use a simplified, informal
+version of [our RFC process][rfcs], where we have design discussions inside the
+Gloo issues tracker.
```
rylev: Nit: It is (and should be) ambiguous when a design spec is warranted and when a direct PR is more appropriate. Should we encourage those who are unsure to open an issue?

Pauan: Yes, as a general rule issues should be opened first.

It's only after there is some consensus about the design that a PR can be opened.

fitzgen: FWIW, this text is largely addressing "new designs/APIs" where I think we practically always want the design discussion first before any PRs.

For bug fixes and other semver compatible changes that don't increase the API surface, I do think we can usually jump straight to a PR, but it also isn't really addressed by this documentation. We could probably add an aside about this somewhere. I'll look into it.

fitzgen: Added a note about purely internal bug fixes in the latest commit.

fitzgen:
Ok, once CI comes back green I'll merge this, since it seems like we all have general agreement that this is the direction we want to move in. As we work out the kinks in this workflow we can definitely adjust things and send follow up PRs as necessary.

Thanks everyone!

Issue #64, opened 2019-04-01
Get mdbook building in CI (and publishing to gh-pages for master commits)


fitzgen:
We should also do the link checking support that mdbook has.

wasm-bindgen does this in CI, so we can probably crib a bunch of stuff from there: https://github.com/rustwasm/wasm-bindgen/blob/master/azure-pipelines.yml#L242-L254

cc @David-OConnor @rylev

PR #71, opened 2019-04-07
Build mdbook in CI

ctjhoa:
This is the first step of #64

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -40,3 +40,10 @@ jobs:
       - template: .ci/wasm-pack-test-headless.yml
         parameters:
           browser_flag: --chrome
+  - job: build_book
+    condition: |
+      eq(variables['Build.Reason'], 'PullRequest')
+    displayName: "Doc - build the book"
+    steps:
+      - template: .ci/install-mdbook.yml
+      - script: (cd guide && mdbook build)
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,8 +1,11 @@
+trigger: ["master"]
 pr: ["master"]
 
 jobs:
   - job: test_all_native
     displayName: "Native Tests"
+    condition: |
+      eq(variables['Build.Reason'], 'PullRequest')
     pool:
       vmImage: 'ubuntu-16.04'
     steps:
@@ -15,6 +18,8 @@ jobs:
         displayName: Run cargo check and test
   - job: test_headless_firefox
     displayName: "Headless Browser Tests (Firefox)"
+    condition: |
+      eq(variables['Build.Reason'], 'PullRequest')
     steps:
       - template: .ci/install-rust.yml
         parameters:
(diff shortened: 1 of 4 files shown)


rylev:
@ctjhoa are you planning on also publishing the mdbook to GitHub pages? Tokio is already doing this so we can just follow their example: https://github.com/tokio-rs/tokio/blob/master/ci/azure-deploy-docs.yml#L20-L38

ctjhoa:
> @ctjhoa are you planning on also publishing the mdbook to GitHub pages?
I'm not sure how to proceed as it require a GITHUB_TOKEN which I suppose should be generated by someone from the team. Am I right ?

rylev:
@ctjhoa yea we'll need to create a deploy key. I don't have permission either but we can work with @fitzgen to create a key and upload in the right place. There's some pretty good docs [here](https://cloudblogs.microsoft.com/opensource/2019/04/05/publishing-github-pages-from-azure-pipelines/?WT.mc_id=twitter-social-rylevick). I'm happy to help :-D

ctjhoa:
I've done some rewriting but it should be OK now.
So I've added:
- A job which build the book on PR (I don't know if it's relevant, maybe a markdown linter should be better)
- A job which publish the book on commit into master

ctjhoa:
An alternative on PR could be to run `mdbook test` which tests that a book's Rust code samples compile.

fitzgen:
This looks good to me, but I also don't really know a ton about what's going on -- did you want to review this PR one last time @rylev?

> An alternative on PR could be to run `mdbook test` which tests that a book's Rust code samples compile.

Last I checked there were issues with using external crates in `mdbook test`. It also means that every example would have to have fully formed code, which can be annoying at times.

.azure-pipelines.yml
```diff
@@ -33,3 +40,19 @@ jobs:
...
         parameters:
           browser_flag: --chrome
+  - job: build_book
+    condition: |
+      eq(variables['Build.Reason'], 'PullRequest')
+    displayName: "Doc - build the book"
+    steps:
+      - template: .ci/install-mdbook.yml
+      - script: (cd guide && mdbook build)
+  - job: deploy_book
```
rylev: why are build and deploy two separate jobs instead of just being steps that happen one after another?

.azure-pipelines.yml
```diff
@@ -33,3 +40,19 @@ jobs:
...
+    displayName: "Doc - deploy the book"
+    condition: |
+      and(not(eq(variables['Build.Reason'], 'PullRequest')),
+          eq(variables['Build.SourceBranch'], 'refs/heads/master'))
```
rylev: You should only need this line and I belive you can get rid of all `variables['Build.Reason'], 'PullRequest')`

ctjhoa: This condition can from [this doc](https://cloudblogs.microsoft.com/opensource/2019/04/05/publishing-github-pages-from-azure-pipelines/?WT.mc_id=twitter-social-rylevick):

> That condition is a little wild as well. You can read it like a prefix-notation functional language (or an Excel formula, if you prefer): “Run this step only if the variable Build.Reason is NOT ‘PullRequest’ and the variable Build.SourceBranch is ‘master’.”

.azure-pipelines.yml
ctjhoa: They don't get the same triggers.
On PR we only check if we can build the book.
On changes occurring on master we redeploy the book.

As I said, we could get rid of the first one if it's not useful.

hamza1311:
Invalidated by #132 and #140

PR #74, opened 2019-04-10
Add docs on versioning and publishing releases to CONTRIBUTING.md

### LINKED ISSUE #66: Questions about versioning Gloo
I'd like to ask two questions related to versioning:

1. How strict should we be with backwards compatibility in Gloo?
2. What is the relationship between the umbrella `gloo` crate's version and each `gloo_whatever` crate's version?

*Note: I'm assuming that whatever the answers, we are following cargo's interpretation of semver.*

## Backwards Compatibility

On one end of the spectrum is never breaking backwards compatibility. This has the advantage of making upgrades to the latest version of crates easy for users. The disadvantage is that our APIs, once published, are set in stone: we can only add new APIs, can't ever fix flaws in existing APIs.

The Rust `std` library takes this to an extreme, where backwards compatibility is guaranteed indefinitely. `wasm-bindgen` is a little looser in that we haven't ruled out breaking version bumps, but we are trying pretty hard to avoid them. We will likely have a single breaking version before moving to 1.0, after which we likely won't do breaking changes anymore.

The other end of the spectrum is fixing flawed APIs whenever we notice them and releasing breaking version bumps immediately. The advantages are that we can fix flaws immediately, and don't have to maintain code that we don't believe meets our design standards. The disadvantages are more difficult upgrades for users, so they are less likely to get the latest and greatest bug fixes and perf improvements.

## Versioning the Umbrella Crate and the Individual Crates

Do we want to keep each individual `gloo_whatever` crate's version locked to the same version as the umbrella `gloo` crate or not?

Imagine this scenario: we have `gloo`, `gloo_timers`, and `gloo_events` all at version `0.1.2`:

```
gloo @ 0.1.2
  - gloo_timers @ 0.1.2
  - gloo_events @ 0.1.2
```

Now we fix a bug in `gloo_timers` that causes us to bump its version from `0.1.2` to `0.1.3`. To get the fix in the umbrella `gloo` crate, we have to bump it to `0.1.3` as well. But do we *also* bump `gloo_events` to `0.1.3` or do we leave it at `0.1.2`?

```
gloo @ 0.1.3
  - gloo_timers @ 0.1.3
  - gloo_events @ 0.1.2 or 0.1.3?
```

The advantages of keeping everything in lock step is largely a lighter maintenance burden for ourselves, the authors/maintainers of Gloo. The disadvantage is unnecessary version bumps for toolkit crates. This disadvantage is magnified by breaking version bumps: in the previous example, if it was a breaking bump to `0.2.0` instead of a compatible bump to `0.1.3`, then we could force an unnecessary breaking bump on `gloo_events`, which would be kind of crappy for anyone using that crate by itself without the rest of Gloo.

The advantages and disadvantages of doing the minimum cascading version bumps are basically the inverse of those for keeping everything in lock step. Upgrades are easier for users (particularly those that are using individual crates, and not the umbrella crate). However, the maintenance burden is larger for Gloo maintainers.

*Note that a breaking change in a `gloo_whatever` crate will always cascade as a breaking change in the umbrella `gloo` crate.*

fitzgen:
cc @Pauan @thedodd @alexcrichton 

Fixes #66

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,9 @@ yourself.
   - [Implementation and Feedback Cycle](#implementation-and-feedback-cycle)
     - [Implementation Checklist](#implementation-checklist)
 - [Team Members](#team-members)
+- [Publishing Releases and Versioning](#publishing-releases-and-versioning)
+  - [Versioning](#versioning)
+  - [Publishing Checklist](#publishing-checklist)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -243,3 +243,55 @@ Team members sign off on design proposals and review pull requests to Gloo.
 If you make a handful of significant contributions to Gloo and participate in
 design proposals, then maybe you should be a team member! Think you or someone
 else would be a good fit? Reach out to us!
+
+## Publishing Releases and Versioning
+
+### Versioning
+
+Whenever we bump a `gloo_whatever` utility crate's version, make sure to make a
+corresponding version bump for every crate that transitively depends upon
+`gloo_whatever`. Crates that do not transitively depend on `gloo_whatever`
+should not have their version bumped.
+
+For example, if we start with this dependency graph:
+
+```
+- gloo @ 0.1.2
+  - gloo_foo @ 0.1.2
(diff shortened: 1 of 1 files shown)


thedodd:
LGTM! +1

fitzgen:
Think it just got forgotten!

PR #75, opened 2019-04-15
Improving the RFC process

Pauan:
See https://github.com/rustwasm/gloo/issues/50#issuecomment-481363615 for more context.


yoshuawuyts:
Bit late to reply, but yeah :100:!

PR #76, opened 2019-04-17
Add a mid-level file-rfc

rylev:
[rendered](https://github.com/rustwasm/gloo/blob/ec9c0c376ead79ff690d124c7b59789650dd4ea8/rfcs/001-mid-level-file-api.md)

As we've settled on an RFC process, I'd like to continue #50 here.

### DIFF
--- a/rfcs/001-mid-level-file-api.md
+++ b/rfcs/001-mid-level-file-api.md
@@ -55,12 +55,12 @@ There are two structs that implement this trait: `Blob` and `File`.
 struct Blob { ... }
 
 impl Blob {
-  fn new<T>(content: T) -> Blob
+  fn new<T>(content: Option<T>) -> Blob
     where
         T: std::convert::Into<BlobContents> // We'll look at BlobContents below
     { ... }
 
-  fn new_with_options<T>(content: T, mime_type: String) -> Blob
+  fn new_with_options<T>(content: Option<T>, mime_type: String) -> Blob
     where
         T: std::convert::Into<BlobContents>
     { ... }
--- a/rfcs/001-mid-level-file-api.md
+++ b/rfcs/001-mid-level-file-api.md
@@ -1,0 +1,187 @@
+# Mid-Level File Api
+
+## Summary
+
+This is a mid-level API wrapper around the various File related JavaScript apis. The possibility for a higher level API is left open.
+
+The mid-level file API aims to implement File I/O on top of the raw JS apis found here: https://w3c.github.io/FileAPI/ which includes listing files, creating files, and reading files.
+
+## The API
+
+First we have `FileList` which is simply a list of `File`s. This is an opaque struct that offers a way to iterate over `Files` and get individual files by index:
+
+```rust
+#[derive(Debug, Clone)]
+struct FileList { ... }
+
+impl FileList {
+  fn get(index: usize) -> Option<File> { ... }
+
+  fn len(&self) -> usize { ... }
+
+  fn iter(&self) -> FileListIter { ... }
+
+  fn as_raw(&self) -> &web_sys::FileList { ... }
+}
+
+impl AsRef<web_sys::FileList> for FileList { ... }
+impl From<web_sys::FileList> for FileList { ... }
+impl Index<usize> for FileList {
+    type Output = File;
+    fn index(&self, index: usize) -> &Self::Output { ... }
+}
+```
+
+There is no `FileList::new` since creating a raw `web_sys::FileList` is not possible without going through a `web_sys::HtmlInputElement`.
+
+Next we have a `blob` module that includes the trait `BlobLike`
+```rust
+trait BlobLike {
+    fn size(&self) -> u64 { ... }
+
+    // the mime crate and mimes from blobs both conform to rfc6838
+    #[cfg(feature = "mime")]
+    fn mime_type(&self) -> Result<mime::Mime, mime::FromStrError> { ... }
+
+    fn raw_mime_type(&self) -> String { ... }
+
+    fn as_raw(&self) -> &web_sys::Blob;
+}
+```
+There are two structs that implement this trait: `Blob` and `File`.
+
+```rust
+#[derive(Debug, Clone)]
+struct Blob { ... }
+
+impl Blob {
+  fn new<T>(content: T) -> Blob
+    where
+        T: std::convert::Into<BlobContents> // We'll look at BlobContents below
+    { ... }
+
+  fn new_with_options<T>(content: T, mime_type: Option<String>) -> Blob
+    where
+        T: std::convert::Into<BlobContents>
+    { ... }
+
+  fn slice(&self, start: usize, end: usize) -> Blob { ... }
+}
+
+impl From<web_sys::Blob> for Blob { ... }
+
+
+impl BlobLike for Blob { ... }
+
+#[derive(Debug, Clone)]
+pub struct File { ... }
+
+impl File {
+  fn new<T>(
+        name: String,
+        contents: Option<T>,
+        mime_type: Option<String>,
+        last_modified_date: Option<u64>,
+    ) -> File
+    where
+        T: std::convert::Into<BlobContents>,
+    { ... }
+
+  fn name(&self) -> String { ... }
+
+  fn last_modified_date(&self) -> u64 { ... }
+
+  fn slice(&self, start: usize, end: usize) -> File { ... }
+
+  fn as_blob(self) -> Blob { ... }
+}
+
+impl BlobLike for File { ... }
+```
+
+Both Blob and File come with builders that allow for creating new Blobs and Files in the builder style and not having to use `new` directly.
+
+`BlobContents` is simply a new-type around `wasm_bindgen::JsValue`s that can be used as the content of `Blob`s and `File`s:
+
+```rust
+#[derive(Debug, Clone)]
+pub struct BlobContents {
+    inner: wasm_bindgen::JsValue,
+}
+```
+
+There are there conversions from types into `BlobContents` only for the types that make sense:
+
+```rust
+impl std::convert::Into<BlobContents> for &str
+impl std::convert::Into<BlobContents> for &[u8]
+impl std::convert::Into<BlobContents> for Blob
+impl std::convert::Into<BlobContents> for js_sys::ArrayBuffer
+```
+
+Lastly there's the `FileReader` which allows reading from BlobLike objects. We'll have two implementations of this, one based on callbacks and the other based on futures.
+
+The callbacks implementation has three categories of callbacks: start, progress and read. Start and progress callbacks are directly related to the `onloadstart` and `onprogress` callbacks on `web_sys::FileReader`. The read variety of callbacks, are a combination of `onload`, `onloadend`, `onloaderror`, and `onabort`. The callback receives a result which is an error if the underlying read was aborted or errored.
+
+The futures implementation likewise exposes success, error, and abort through the fact that futures are `Result`-like. Progress events are exposed as a stream. In the future, we may expose the entire lifecycle of a read through a stream.
+
+```rust
+mod callbacks {
+  #[derive(Debug)]
+  pub struct FileReader { ... }
+
+  impl FileReader {
+    fn new() -> FileReader { ... }
+
+    fn read_as_string<F>(self, blob: &impl BlobLike, callback: F)
+        where F: FnOnce(Result<String, FileReadError>) { ... };
+
+    fn read_as_data_url<F>(self, blob: &impl BlobLike, callback: F)
+        where F: FnOnce(Result<String, FileReadError>) { ... };
+
+    fn read_as_array_buffer<F>(self, blob: &impl BlobLike, callback: F)
+        where F: FnOnce(Result<&web_sys::ArrayBuffer, FileReadError>) { ... };
+
+    fn on_progress<F>(&mut self, callback: F)
+      where F: FnMut(ProgressEvent) + 'static { ... }
+
+    fn on_load_start<F>(&mut self, callback: F)
+      where F: FnOnce(LoadStartEvent) + 'static { ... }
+  }
+}
+
+mod futures {
+  #[derive(Debug)]
+  pub struct FileReader { ... }
+
+  impl FileReader {
+    fn new() -> FileReader { ... }
+
+    fn read_as_string(self, blob: &impl BlobLike) -> ReadAsString { ... }
+
+    fn read_as_data_url(self, blob: &impl BlobLike) -> ReadAsDataUrl { ... }
+
+    fn read_as_array_buffer(self, blob: &impl BlobLike) -> ReadAsArrayBuffer { ... }
+
+    fn on_progress(&self) -> OnProgressStream { }
+
+    fn on_load_start(&self) -> OnLoadStartFuture { }
+  }
+
+  pub struct ReadAsString { ... }
+  impl Future for ReadAsString {
+      type Item = String;
+      type Error = FileReadError;
+      ...
+  }
+
+  // Make sure that dropping the Future properly aborts the reading
+  impl std::ops::Drop for ReadAsString {
+      fn drop(&mut self) {
+          if self.inner.ready_state() < 2 {
+              self.inner.abort();
+          }
+      }
+  }
+}
+```
--- a/rfcs/001-mid-level-file-api.md
+++ b/rfcs/001-mid-level-file-api.md
@@ -15,9 +15,9 @@ First we have `FileList` which is simply a list of `File`s. This is an opaque st
 struct FileList { ... }
 
 impl FileList {
-  fn get(index: usize) -> Option<File> { ... }
+  fn get(index: u32) -> Option<File> { ... }
 
-  fn len(&self) -> usize { ... }
+  fn len(&self) -> u32 { ... }
 
   fn iter(&self) -> FileListIter { ... }
 
@@ -34,7 +34,7 @@ impl Index<usize> for FileList {
 
 There is no `FileList::new` since creating a raw `web_sys::FileList` is not possible without going through a `web_sys::HtmlInputElement`.
 
-Next we have a `blob` module that includes the trait `BlobLike`
+Next we the trait `BlobLike`
 ```rust
 trait BlobLike {
     fn size(&self) -> u64 { ... }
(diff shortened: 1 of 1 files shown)


rylev:
I'll wait for @Pauan's feedback and approval before submitting an implementation PR.

rylev:
@yoshuawuyts Having futures in the same API feels right to me, because It still maps very closely to the JS API. 

I don't have strong feelings about this though so if the group wants to be extra conservative and only do the callback based API at first I'm fine with that. We could later open a new proposal that would extend the mid-level API with futures OR create a new high level API with futures.

yoshuawuyts:
> We could later open a new proposal that would extend the mid-level API with futures OR create a new high level API with futures.

Yeah, I'm definitely feeling like having a high-level API on top of this would be a great fit. For example `slice` could be abstracted through [Index](https://doc.rust-lang.org/std/ops/trait.Index.html) / [IndexMut](https://doc.rust-lang.org/std/ops/trait.IndexMut.html) to make it similar to Rust slices.

Also [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html), [Add](https://doc.rust-lang.org/std/ops/trait.Add.html) and other might have a place. I feel Futures provide a similar high-level interface on top of the mid-level API.

But yeah, don't feel too strongly about this. Just curious how we're envisioning we slice this up (:

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+impl From<web_sys::FileList> for FileList { ... }
+impl Index<usize> for FileList {
+    type Output = File;
+    fn index(&self, index: usize) -> &Self::Output { ... }
+}
+```
+
+There is no `FileList::new` since creating a raw `web_sys::FileList` is not possible without going through a `web_sys::HtmlInputElement`.
+
+Next we have a `blob` module that includes the trait `BlobLike`
```
Pauan: Why is this put into a sub-module, but things like `FileList` aren't?

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+```rust
+trait BlobLike {
+    fn size(&self) -> u64 { ... }
```
Pauan: This definitely shouldn't be `u64`, since `u64` cannot be sent to/from JS (it causes an exception).

So it either needs to be `f64` or `u32`. I'm not sure which one.

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+This is a mid-level API wrapper around the various File related JavaScript apis. The possibility for a higher level API is left open.
+
+The mid-level file API aims to implement File I/O on top of the raw JS apis found here: https://w3c.github.io/FileAPI/ which includes listing files, creating files, and reading files.
+
+## The API
+
+First we have `FileList` which is simply a list of `File`s. This is an opaque struct that offers a way to iterate over `Files` and get individual files by index:
+
+```rust
+#[derive(Debug, Clone)]
```
Pauan: All of these structs should probably also impl `PartialEq` and `Eq`.

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+struct FileList { ... }
+
+impl FileList {
+  fn get(index: usize) -> Option<File> { ... }
```
Pauan: `get` and `len` need to be `u32`: https://w3c.github.io/FileAPI/#dfn-filelist

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+#[derive(Debug, Clone)]
+struct Blob { ... }
+
+impl Blob {
+  fn new<T>(content: T) -> Blob
+    where
+        T: std::convert::Into<BlobContents> // We'll look at BlobContents below
+    { ... }
+
+  fn new_with_options<T>(content: T, mime_type: Option<String>) -> Blob
```
Pauan: Why is this `Option`? If somebody didn't want to specify options, they'd just use `Blob::new` instead.

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+    where
+        T: std::convert::Into<BlobContents>
+    { ... }
+
+  fn slice(&self, start: usize, end: usize) -> Blob { ... }
```
Pauan: As above, this needs to be `f64` or `u32`

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+impl From<web_sys::Blob> for Blob { ... }
+
+
+impl BlobLike for Blob { ... }
+
+#[derive(Debug, Clone)]
+pub struct File { ... }
+
+impl File {
+  fn new<T>(
```
Pauan: Since `FileBuilder` exists, why do we need this method?

Since `Blob::new` is so simple, why do we need `BlobBuilder`?

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+        mime_type: Option<String>,
+        last_modified_date: Option<u64>,
+    ) -> File
+    where
+        T: std::convert::Into<BlobContents>,
+    { ... }
+
+  fn name(&self) -> String { ... }
+
+  fn last_modified_date(&self) -> u64 { ... }
```
Pauan: This needs to be `f64`

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+
+```rust
+mod callbacks {
+  #[derive(Debug)]
+  pub struct FileReader { ... }
+
+  impl FileReader {
+    fn new() -> FileReader { ... }
+
+    fn read_as_string<F>(self, blob: &impl BlobLike, callback: F)
```
Pauan: Why do these take `&impl BlobLike` and not `A where A: BlobLike`?

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+        T: std::convert::Into<BlobContents>,
+    { ... }
+
+  fn name(&self) -> String { ... }
+
+  fn last_modified_date(&self) -> u64 { ... }
+
+  fn slice(&self, start: usize, end: usize) -> File { ... }
+
+  fn as_blob(self) -> Blob { ... }
```
Pauan: This should accept `&self`

rfcs/001-mid-level-file-api.md
fitzgen: `impl Trait` is generalyl the preferred short hand unless you need to reference the type parameter in another bound. They should be equivalent, either way.

rfcs/001-mid-level-file-api.md
fitzgen: This is the Rust-facing API, not what is transferred at the ABI level. It would be perfectly valid to transfer the size as an `f64` at the ABI level and then expose it to Rust as a `u64`. This gives us more bits than a `u32` and also lets us ignore silliness about fractional bytes, all while working within the ABI's contstraints.

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Summary
+
+This is a mid-level API wrapper around the various File related JavaScript apis. The possibility for a higher level API is left open.
+
+The mid-level file API aims to implement File I/O on top of the raw JS apis found here: https://w3c.github.io/FileAPI/ which includes listing files, creating files, and reading files.
+
+## The API
+
+First we have `FileList` which is simply a list of `File`s. This is an opaque struct that offers a way to iterate over `Files` and get individual files by index:
```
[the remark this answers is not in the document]
rylev: This would require iterating the internal `js_sys::FileList` and collecting the files into a `Vec`. Are we sure we want the user to pay this price each time a `FileList` is constructed? I personally think an `into_vec` API is more appropriate since it conveys that there will be an additional allocation.

In any effect, it's possible to add these in a backward compatible way. Can we push this off to another PR/RFC?

rfcs/001-mid-level-file-api.md
rylev: I included new since having some sort of `new` function is fairly common even if there's a builder. I included the `BlobBuilder` just for consistency's sake. I'm fine with removing both. We can always add them back.

rfcs/001-mid-level-file-api.md
rylev: `web_sys::FileList` doesn't implement either `PartialEq` or `Eq` so we can't derive it. What would a correct impl for `FileList`?

rfcs/001-mid-level-file-api.md
rylev: Not sure what to do here. `u64` is a possibility for `start` and `end` if we just use `min(start,  f64::MAX)`.

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,194 @@
...
+}
+```
+There are two structs that implement this trait: `Blob` and `File`.
+
+```rust
+#[derive(Debug, Clone)]
+struct Blob { ... }
+
+impl Blob {
+  fn new<T>(content: Option<T>) -> Blob
```
rylev: I moved this to option, because I think it better represents that Blobs can be empty. We could require just `T` but then every time you'd want to create an empty blob, you'd have to do provide `&Vec::new()`. Maybe that's ok?

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,194 @@
...
+        mime_type: Option<String>,
+        last_modified_date: Option<u64>,
+  ) -> File
+    where
+        T: std::convert::Into<BlobContents>,
+    { ... }
+
+  fn name(&self) -> String { ... }
+
+  fn last_modified_since_epoch(&self) -> Duration { ... }
```
rylev: I renamed this method and changed its return type. Thoughts?

rfcs/001-mid-level-file-api.md
Pauan: Ah, right, I had confused it with `&dyn BlobLike`.

I still think it should use generics though, so it can be used with turbofish.

rfcs/001-mid-level-file-api.md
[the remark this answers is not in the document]
Pauan: I don't think it's possible to add it in a backwards compatible way, since a `Vec` would have a different API.

And since (apparently) `FileList` is being phased out in favor of a normal JS `Array`, I think `Vec` is quite appropriate.

In fact, we may even want to skip the `FileList` type entirely and just return `Vec<File>`.

That also makes it pretty clear that there is an allocation (since `Vec` obviously allocates).

rfcs/001-mid-level-file-api.md
Pauan: That's an interesting solution, I hadn't thought of that.

I think we need to be a little careful around the limits with regard to `f64`, and it needs to documented, but it seems doable.

rfcs/001-mid-level-file-api.md
Pauan: They will implement them in the future though. So I guess we can defer this until then.

(The correct impl would just do the same thing JS does, using `===` or `Object.is` to compare by reference).

rfcs/001-mid-level-file-api.md
Pauan: After @fitzgen 's explanation, I'm now on board with `u64` (still not `usize` though).

However, I don't think it should silently saturate, it should panic if it goes out of range.

Also the maximum needs to be `9007199254740991`, not `f64::MAX`.

rfcs/001-mid-level-file-api.md
Pauan: Is it common? I've only seen a clean split between the builder and the final object. Do you have any examples of that?

rfcs/001-mid-level-file-api.md
Pauan: There's a big difference between "empty" (which would be an empty `Vec`) and "nothingness" (which would be `None`).

Since `Blob` can be called with many different types (`Array`, `ArrayBuffer`, other `Blob`s, `String`, etc.) what would it mean for a blob to be "empty"? Would it be an empty string? An empty array?

Since it's mandatory to pass in an argument to the `Blob` constructor, I don't think using `Option` makes sense here.

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,194 @@
...
+
+  fn slice(&self, start: u64, end: u64) -> File { ... }
```
Pauan: Any reason `slice` isn't defined on `BlobLike`?

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,194 @@
...
+
+  fn as_blob(&self) -> Blob { ... }
```
Pauan: Is this needed, since `BlobLike` defines `as_raw`?

rfcs/001-mid-level-file-api.md
Pauan: I think it's okay. The nanosecond precision isn't necessary, but I guess it doesn't hurt either. It is certainly more "Rusty".

Pauan:
(By the way, sorry for the delay on this, I recently caught a nasty cold)

rfcs/001-mid-level-file-api.md
Pauan: Ah, I see the difference. This returns a `Blob`, whereas `as_raw` returns a `web_sys::Blob`.

I'm still not a big fan of this method though. It feels too much like it should be something else, like `Deref<Target = Blob>`, or `AsRef`, or `Into`, or...

rfcs/001-mid-level-file-api.md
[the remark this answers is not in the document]
rylev: @Pauan Can you explain how the API would need to be different? I'm sorry, I don't really get the reason for this given the cost users would have to pay.

rfcs/001-mid-level-file-api.md
rylev: I agree. Ill make the change.

rylev:
@Pauan @fitzgen I've made some final changes including removing the builders for File and Blob since the constructors are now now that unwieldy. 

I'd like to merge this. I know there's one concern not fully addressed from @Pauan, but I think the proposal is in a good state and ready to be implemented. Thoughts?

fitzgen:
I'm happy with it. Let's address the last outstanding concern and get this merged!

rfcs/001-mid-level-file-api.md
fitzgen: `AsRef` and `Deref` don't work because this isn't returning a reference.

Since the conversion's ownership is from borrowed to owned, it should probably be called `to_blob` to follow [the C-CONV guidelines](https://rust-lang-nursery.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv).

fitzgen: @rylev @Pauan how does renaming to `to_blob` sound?

rylev: That sounds fine to me.

rfcs/001-mid-level-file-api.md
```diff
@@ -0,0 +1,190 @@
...
+}
+
+mod futures {
+  #[derive(Debug)]
+  pub struct FileReader { ... }
+
+  impl FileReader {
+    fn new() -> FileReader { ... }
+
+    fn read_as_string(self, blob: &impl BlobLike) -> ReadAsString { ... }
```
rylev: @fitzgen These should also probably be called `read_to_*`. Thoughts?

fitzgen: Sounds good to me!

rylev:
@Pauan @fitzgen shall we merge this?

fitzgen:
> @Pauan @fitzgen shall we merge this?

I'm :+1:, let's just wait for @Pauan's response before merging.

rfcs/001-mid-level-file-api.md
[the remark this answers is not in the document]
Pauan: Basically, rather than having a method which returns a `FileList`, it would instead return `Vec<File>`.

That means we don't actually need `FileList` at all.

The cost is quite small, but the benefits are significant: things are drastically simplified, and users get to use the slice methods for free.

rfcs/001-mid-level-file-api.md
Pauan: Yeah, I like `to_blob` better.

rylev:
@Pauan I'm not convinced. If the user wants a `Vec<File>` they can call `iter().collect::<Vec<File>>()`. If we think that's too much, we can always include a `to_vec()` method that does that for the user. I think the proposal as is does a good job of walking the line between feeling like Rust and still mirroring the JS API fairly closely. Perhaps we'll want to have a higher level API that goes even further?

Pauan:
Yes of course the user can do that... but the point is to simplify things. Not needing to have the `FileList` type (or any of the various methods), being able to use slice methods for free. For the user it's great!

It's less things to learn, less things to worry about. It's also extremely Rusty (since `Vec` is one of the core Rust data types).

And as the W3C page says, they're deprecating `FileList` and moving toward a JS `Array` (which maps directly to `Vec` in Rust). So it's also quite natural from that perspective as well.

rylev:
@Pauan Do you propose just doing `impl From<web_sys::FileList> for Vec<File>` for now?  I'd be interested in @fitzgen opinion here. Perhaps he can break the deadlock?

Pauan:
Yes, that's basically my proposal. Though we haven't yet decided on how this will interact with `<input>` (since that's a separate proposal).

fitzgen:
### `gloo_files::FileList`

#### Pros

* Could be made zero cost: if we add a `'a` lifetime parameter so it is `gloo_files::FileList<'a>` and make it wrap a `anyref` stack-allocated `&'a web_sys::FileList`, then there are no heap allocations involved. `Vec<gloo_files::File>` requires heap allocations.

#### Cons

* Additional type to learn in the API.

* Spec is moving away from `web_sys::FileList` and moving towards JS array. To get the no-implicit-heap-allocation version in this world, `gloo_files::FileList<'a>` would have to wrap a `&'a js_sys::Array`.

### `Vec<gloo_files::File>`

#### Pros

* Fewer types in the API to learn.

* Conceptually matches where the spec is heading.

#### Cons

* Requires heap allocation.

### Conclusions

In the case of a single file in the file list, it is *nice* to avoid the `Vec`'s heap allocation. But it also doesn't seem like a *hard requirement* in this case, since it is just a single heap allocation that shouldn't ever be part of a loop or anything like that. It seems to me like this particular allocation doesn't really matter.

Therefore I think we should go with `Vec<gloo_files::File>` and not `gloo_files::FileList`.

But mostly I think we would be just fine with either design, it isn't a big deal, and we are bike shedding at this point. Its been 5 weeks since this API was first proposed, and while it has certainly improved from the initial proposal, at this point I think we gain more by just merging than by deliberating further.

rylev:
@Pauan @fitzgen updated. Let's merge this.

PR #78, opened 2019-04-19
Derive Debug impls

lnicola:
Not sure this is an improvement, but hey, at least it fixes a TODO.

I also sneaked in a `Cargo.toml` change to avoid a warning.

### DIFF
--- a/crates/console-timer/Cargo.toml
+++ b/crates/console-timer/Cargo.toml
@@ -11,5 +11,5 @@ features = [
 ]
 
 [dev-dependencies]
-wasm-bindgen-test = "0.2.40"
+wasm-bindgen-test = "0.2.43"
 gloo-timers = { version = "0.1.0", path = "../timers" }
--- a/crates/events/Cargo.toml
+++ b/crates/events/Cargo.toml
@@ -21,6 +21,7 @@ futures = "0.1.25"
 wasm-bindgen-test = "0.2.37"
 
 [dev-dependencies.web-sys]
+version = "0.3.14"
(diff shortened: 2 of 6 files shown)


Pauan:
One thing I would suggest is to increase the version of `wasm-bindgen` to `0.2.41`, since that's the version that introduced `Debug` for `Closure`.

lnicola:
Filed https://github.com/rustwasm/wasm-bindgen/pull/1477 for the `JsFuture` impl.

I'm actually not sure that this PR is a good change, since it makes the debug output more crowded while saving a couple of lines of code.

Pauan:
[`Debug` isn't really intended to be "clean"](https://doc.rust-lang.org/nightly/std/fmt/index.html#fmtdisplay-vs-fmtdebug), it's intended for situations where you're debugging problems.

So it's usually better to err on the side of providing more information.

lnicola:
> So it's usually better to err on the side of providing more information.

Sure, but if it's always going to print a `JsFuture { ... }` field, it doesn't bring anything of value. I see it like [`Sender`](https://doc.rust-lang.org/src/std/sync/mpsc/mod.rs.html#915).

Pauan:
It does provide value though: it lets you know that it's implemented with `JsFuture` (so you can then go and look up the docs/implementation for that).

Also, not using `derive(Debug)` is a refactoring hazard, since somebody might forget to update the `Debug` impl when making changes to the type.

lnicola:
This is now ready, methinks.

Issue #5, opened 2019-02-26
Web GL


fitzgen:
Can we make something like `glium` for Web GL? Does it make sense to jump straight there, or should we build an incremental Rustier version of the Web GL APIs before trying to make an Opinionated(tm) library?

grovesNL:
IMO it might be useful to have unified OpenGL, OpenGL ES, and WebGL bindings, similar to the unified bindings Emscripten provides. I already started working on this in https://github.com/grovesNL/glow for use in gfx-hal's GL backend (gfx-backend-gl) and I'll continue to expand it as I work through WebGL support for gfx-hal.

With this approach, projects like gfx-hal, glium etc. could use and share these unified bindings instead.

asny:
After reading the [Gloo update](https://rustwasm.github.io/2019/03/26/gloo-onion-layers.html), I will suggest to start with a minimalistic WebGL and WebGL2 wrapper which have three main goals:
* Wrap the web-sys functions in nice Rust-y types and interfaces and with a 1-1 correspondence to the WebGL API.
* Elegantly handle as many error cases as possible without any significant overhead.
* Ensure performance wherever it is possible for example by ensuring the state is not changed if the specified state is the same as the current.

.. So basically the same purpose as Glium but without the simplification of the interface. I think it is important to have the thin Rust-y wrapper on top of WebGL as the innermost layer of the onion and when that is done, there is plenty of room to add outer layers. This could be a simplified and safer interface or the combined OpenGL+WebGL API suggested by @grovesNL.

Please don’t hesitate to disagree with some or any of my opinions :-)

David-OConnor:
What do y'all thinking of mimicking syntax from GFX/Vulkan when able?

I agree that the first step should be a thin wrapper. For later high-level wrappers, it would be nice to include matrix-manipulation tools. Using [cgmath](https://github.com/rustgd/cgmath) as a dependency is one option, or more generally, [ndarray](https://github.com/rust-ndarray/ndarray). Or creating our own. Perhaps this is required for easy conversion to JS TypedArrays?

yoshuawuyts:
I have no opinions here, other than that [regl.party](http://regl.party) has proven to be a useful model to build longer-lasting WebGL applications in JS. Perhaps it's something to draw inspiration from when we eventually build the higher-level layers.

asny:
Here’s a more detailed proposal of the basic WebGL crate after request by @fitzgen.

## Summary

[...]

PR #79, opened 2019-05-01
[RFC] WebGL

asny:
Continuing from issue [#5](https://github.com/rustwasm/gloo/issues/5)

### DIFF
--- a/rfcs/webgl.md
+++ b/rfcs/webgl.md
@@ -1,0 +1,145 @@
+## Summary
+
+Safe, rusty wrapper for the web-sys WebGL functionality.
+
+What is included?
+* Only standard rust types in the interface.
+* Type safe interface.
+* Error handling by returning results from all functions that can cause a runtime error.
+* Simplification of the interface wherever it does not limit the functionality. (note that this opinion differ a bit from my previous post, I think it is not possible to have a 1-1 correspondence to the WebGL API, but it should be as close as possible)
+
+What is not included?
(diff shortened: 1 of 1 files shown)


derekdreery:
Awesome work. I'm not really placed to offer an opinion on the proposal as a whole, but I have a couple of questions:

 1. If there are some combinations of parameters that are not allowed, can we create types that represent the allowed subset using sum and product types (enums and structs)?
 2. Do we want to support WebGL rather than WebGPU, since people using wasm are going to have an up-to-date browser, and WebGPU is designed to replace WebGL?

Pauan:
@derekdreery I think the problem is that WebGPU isn't even remotely close to ready (it's not standardized yet, the design is still in flux, it currently [only runs on Mac OS X Chrome Canary](https://doc.babylonjs.com/extensions/webgpu), *and* you have to enable an experimental flag, *and* it's missing a ton of features, *and* it's going to be buggy).

But yeah, in the long term we'll want to replace WebGL with WebGPU.

hamza1311:
Hey @ansy, can you create an issue for this RFC. See #136 for details

asny:
Hi @hamza1311 , it is already there #5

hamza1311:
> Hi @hamza1311 , it is already there #5 

You mentioned that it's a continuation. I haven't read *both* so can you confirm that it's all the information is there?

hamza1311:
Just read it, seems like the issue does have the information. Closing this in favor of #5

PR #83, opened 2019-05-17
Add implementation for rfc-1 (files)

rylev:
Implementation for [rfc-1](https://github.com/rustwasm/gloo/blob/master/rfcs/001-mid-level-file-api.md).

This implementation is not complete. Left to do:
* [ ] Finalize design for errors
* [ ] Implement event based callbacks
* [ ] Tests
* [ ] Documentation

### DIFF
--- a/crates/file/src/blob.rs
+++ b/crates/file/src/blob.rs
@@ -36,41 +36,46 @@ impl BlobContents {
     }
 }
 
-impl std::convert::Into<wasm_bindgen::JsValue> for BlobContents {
-    fn into(self) -> wasm_bindgen::JsValue {
-        self.inner
+impl std::convert::From<BlobContents> for wasm_bindgen::JsValue {
+    fn from(blob_contents: BlobContents) -> wasm_bindgen::JsValue {
+        blob_contents.inner
     }
 }
 
-impl std::convert::Into<BlobContents> for &str {
-    fn into(self) -> BlobContents {
-        BlobContents::from_raw(wasm_bindgen::JsValue::from_str(self))
+impl std::convert::From<&str> for BlobContents {
+    fn from(str: &str) -> Self {
+        BlobContents::from_raw(wasm_bindgen::JsValue::from_str(str))
     }
 }
 
-impl std::convert::Into<BlobContents> for &[u8] {
-    fn into(self) -> BlobContents {
-        let array = unsafe { js_sys::Uint8Array::view(self) };
-        let array_clone = js_sys::Uint8Array::new(&array).buffer();
-        BlobContents::from_raw(array_clone.into())
+impl std::convert::From<&[u8]> for BlobContents {
+    fn from(buffer: &[u8]) -> Self {
+        let array = unsafe { js_sys::Uint8Array::view(buffer) };
+        BlobContents::from_raw(js_sys::Uint8Array::new(&array).into())
     }
 }
 
-impl std::convert::Into<BlobContents> for Blob {
-    fn into(self) -> BlobContents {
-        BlobContents::from_raw(self.inner.into())
+impl std::convert::From<Blob> for BlobContents {
+    fn from(blob: Blob) -> Self {
+        BlobContents::from_raw(blob.inner.into())
     }
 }
 
-impl std::convert::Into<BlobContents> for web_sys::Blob {
-    fn into(self) -> BlobContents {
-        BlobContents::from_raw(self.into())
+impl std::convert::From<File> for BlobContents {
+    fn from(file: File) -> Self {
+        BlobContents::from_raw(file.inner.into())
     }
 }
 
-impl std::convert::Into<BlobContents> for js_sys::ArrayBuffer {
-    fn into(self) -> BlobContents {
-        BlobContents::from_raw(self.into())
+impl std::convert::From<web_sys::Blob> for BlobContents {
+    fn from(blob: web_sys::Blob) -> Self {
+        BlobContents::from_raw(blob.into())
+    }
+}
+
+impl std::convert::From<js_sys::ArrayBuffer> for BlobContents {
+    fn from(buffer: js_sys::ArrayBuffer) -> Self {
+        BlobContents::from_raw(buffer.into())
     }
 }
 
@@ -125,11 +130,17 @@ impl BlobLike for File {
             .as_raw()
             .slice_with_f64_and_f64(start as f64, end as f64)
             .unwrap_throw();
+        let raw_mime_type = self.raw_mime_type();
+        let mime_type = if raw_mime_type == "" {
+            None
+        } else {
+            Some(raw_mime_type)
+        };
 
         File::new_with_options(
             self.name(),
             blob,
-            None,
+            mime_type,
             Some(self.last_modified_since_epoch()),
         )
     }
@@ -165,13 +176,8 @@ impl File {
             options.type_(&mime_type);
         }
         if let Some(last_modified) = last_modified_since_epoch {
-            let (millis, did_overflow_seconds) = last_modified.as_secs().overflowing_mul(1000);
-            let (millis, did_overflow_millis) =
-                millis.overflowing_add(last_modified.subsec_millis() as u64);
-            assert!(
-                !did_overflow_seconds && !did_overflow_millis,
-                "last modified since epoch duration overflowed"
-            );
+            let millis = last_modified.as_millis();
+            // TODO: max check
 
             options.last_modified(millis as f64);
         }
--- a/crates/file/src/blob.rs
+++ b/crates/file/src/blob.rs
@@ -113,6 +113,8 @@ impl BlobLike for Blob {
     }
 
     fn slice(&self, start: u64, end: u64) -> Self {
+        ensure_representable(start);
+        ensure_representable(end);
         Blob::from_raw(
             self.as_raw()
                 .slice_with_f64_and_f64(start as f64, end as f64)
@@ -126,6 +128,8 @@ impl BlobLike for File {
     }
 
     fn slice(&self, start: u64, end: u64) -> Self {
+        ensure_representable(start);
+        ensure_representable(end);
         let blob = self
             .as_raw()
             .slice_with_f64_and_f64(start as f64, end as f64)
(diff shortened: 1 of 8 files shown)


crates/file/src/file_list.rs
```diff
@@ -0,0 +1,23 @@
+use crate::blob::File;
+
+pub struct FileList {
+    inner: Vec<File>,
+}
+
+impl std::convert::Into<FileList> for web_sys::FileList {
```
fitzgen: We should implement `From<web_sys::FileList> for FileList` and then we get the `Into` impl for free as well because there is a blanket `impl<T, U> Into<T> for U where T: From<U>`.

fitzgen: Same for all the other `Into` impls. Generally, you never want to implement `Into` manually, only `From`.

Also, `Into` and `From` are in the prelude, so you don't need to fully qualify them.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    impl FileReader {
+        pub fn new() -> FileReader {
+            FileReader {
+                inner: web_sys::FileReader::new().unwrap_throw(),
+            }
+        }
+
+        pub fn read_as_string(self, blob: &impl BlobLike) -> ReadAsString {
```
Pauan: I think these should be top-level functions too.

Also, there's a lot of code duplication here. These should be using the `callbacks` functions and just wrapping them in Futures.

That means the `callbacks` functions need to return something which enables them to be dropped/cancelled.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    macro_rules! future_impls {
+       ($($name:ident : $item:ty),*) => ($(
+            pub struct $name {
+                receiver: oneshot::Receiver<$item>,
+                _closure: Closure<FnMut()>,
+                inner: web_sys::FileReader,
+            }
+
+            impl Future for $name {
```
Pauan: I think we should seriously consider using Futures 0.3 instead.

The `Future` trait is already stabilized, and Futures 0.3 will be available on stable soon.

I'm also okay with delaying the `futures` submodule until after Futures 0.3 is available on stable (only releasing the `callbacks` submodule).

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    future_impls! { ReadAsString : String, ReadAsDataUrl: String, ReadAsArrayBuffer: js_sys::ArrayBuffer }
+}
+
+#[derive(Debug)]
+pub enum FileReadError {
+    AbortedEarly,
+    JsError(wasm_bindgen::JsValue),
+}
+impl std::convert::Into<FileReadError> for wasm_bindgen::JsValue {
```
Pauan: This feels extremely broad. Since there are a limited number of errors which can be thrown, we shouldn't allow for converting anything into a `FileReadError`.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+       )*)
+    }
+
+    future_impls! { ReadAsString : String, ReadAsDataUrl: String, ReadAsArrayBuffer: js_sys::ArrayBuffer }
+}
+
+#[derive(Debug)]
+pub enum FileReadError {
+    AbortedEarly,
+    JsError(wasm_bindgen::JsValue),
```
Pauan: This should take an `Error` object. Or even better, we can just list out all of the possible errors that can happen.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+impl std::convert::Into<FileReadError> for wasm_bindgen::JsValue {
+    fn into(self) -> FileReadError {
+        FileReadError::JsError(self)
+    }
+}
+
+impl std::convert::Into<FileReadError> for ::futures::sync::oneshot::Canceled {
```
Pauan: I don't like this impl. It's exposing our internal implementation details.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+    fn into(self) -> FileReadError {
+        FileReadError::AbortedEarly
+    }
+}
+
+impl std::fmt::Display for FileReadError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "error reading file")
```
Pauan: This can probably be a lot nicer: for example printing the `Error` object's `message`.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,248 @@
...
+            convert: ConvertFn,
+            read: ReadFn,
+            callback: CallbackFn,
+        ) where
+            ConvertFn: Fn(wasm_bindgen::JsValue) -> T + 'static,
+            ReadFn: Fn(&web_sys::FileReader, &web_sys::Blob) -> Result<(), wasm_bindgen::JsValue>,
+            CallbackFn: FnOnce(Result<T, FileReadError>) + 'static,
+        {
+            let reader = self.inner.clone();
+            let closure = Closure::once(move || {
```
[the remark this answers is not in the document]
Pauan: (I really should have caught this and the `self` issue during the RFC stage, sorry about that)

crates/file/src/file_list.rs
rylev: I always get this backwards. Will make the change.

crates/file/src/blob.rs
```diff
@@ -0,0 +1,200 @@
...
+        let mut properties = web_sys::BlobPropertyBag::new();
+        properties.type_(&mime_type);
+
+        let parts = js_sys::Array::of1(&content.into().inner);
+        let inner = web_sys::Blob::new_with_u8_array_sequence_and_options(&parts, &properties);
+
+        Blob::from_raw(inner.unwrap_throw())
+    }
+
+    pub fn from_raw(inner: web_sys::Blob) -> Blob {
```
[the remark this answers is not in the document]
rylev: This convention is mentioned in the contributing.md. Perhaps that should be changed?

crates/file/src/blob.rs
```diff
@@ -0,0 +1,200 @@
...
+    ) -> File
+    where
+        T: std::convert::Into<BlobContents>,
+    {
+        let mut options = web_sys::FilePropertyBag::new();
+        if let Some(mime_type) = mime_type {
+            options.type_(&mime_type);
+        }
+        if let Some(last_modified) = last_modified_since_epoch {
+            let (millis, did_overflow_seconds) = last_modified.as_secs().overflowing_mul(1000);
```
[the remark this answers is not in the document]
rylev: Ah! I didn't realize `as_millis` had been stabilized! 🎉

crates/file/src/blob.rs
```diff
@@ -0,0 +1,200 @@
...
+            options.last_modified(millis as f64);
+        }
+        let parts = js_sys::Array::of1(&contents.into().inner);
+        let inner = web_sys::File::new_with_u8_array_sequence_and_options(&parts, &name, &options)
+            .unwrap_throw();
+
+        File::from_raw(inner)
+    }
+
+    pub fn from_raw(inner: web_sys::File) -> File {
```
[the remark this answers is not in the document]
rylev: See reply above.

rylev:
@Pauan I believe I've addressed most of your issues. I've removed the futures based API for now since I think it complicates things and can easily be added in a following PR without breaking anything. Let me know what you think.

Of course, the work items I wrote out above still need to be addressed.

crates/file/src/blob.rs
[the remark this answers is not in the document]
Pauan: I'm undecided. I guess `from_raw` is fine for now.

Pauan:
@rylev I'm in the middle of reviewing it, there's some tricky parts I need to think about.

fitzgen:
> @Pauan @fitzgen thoughts? Is this good enough to be cleaned up with tests and documentation?

In my opinion, definitely yes.

I think we are at the point where it makes more sense to get these docs and tests and merge this PR than it does to nitpick on too many little details right now. Once this is merged and a few other little things, we can cut a 0.1 release of gloo and we can start getting people using it and getting excitement around it and gathering feedback. I think those things would do a lot more for pushing this project forward and getting closer to fulfilling its goals.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum FileReadError {
+    AbortedEarly,
+    JsError(js_sys::Error),
+}
+impl std::convert::From<wasm_bindgen::JsValue> for FileReadError {
```
Pauan: This feels a bit odd: should it take `js_sys::Error` instead?

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,110 @@
...
+        let closure = Closure::once(move || {
+            let result = clone
+                .result()
+                .map(convert)
+                .map_err(std::convert::Into::into);
+            callback(result);
+        });
+        let function = closure.as_ref().dyn_ref().unwrap_throw();
+
+        reader.clone().set_onload(Some(&function));
```
Pauan: This `clone` doesn't seem necessary.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,110 @@
+pub mod callbacks {
+    use super::FileReadError;
+    use crate::blob::BlobLike;
+    use wasm_bindgen::{closure::Closure, JsCast, UnwrapThrowExt};
+
+    #[derive(Debug)]
+    pub struct FileRead {
```
Pauan: I think it sounds a bit better if this is called `FileReader`

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,110 @@
...
+                self.reader.abort();
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum FileReadError {
+    AbortedEarly,
+    JsError(js_sys::Error),
```
Pauan: I'm not 100% sure, but it sounds like [there's only 3 possible errors that can occur](https://w3c.github.io/FileAPI/#file-error-read), so maybe we want to list them out explicitly, rather than using a generic `Error` branch?

derekdreery:
@rylev would you mind if I picked this up and worked on it?

rylev:
@derekdreery I do not mind. I am unfortunately too busy to see this to the end. 😞

hamza1311:
> This PR should be closed as it is superseeded by #83

This PR is #83...

derekdreery:
Ah yep apologies the PR I meant is #110

PR #90, opened 2019-06-30
Annotate trait objects ready for when they become mandatory.

derekdreery:
This is basically a no-op commit that fixes some warnings on latest compiler.


Pauan:
Thanks!

Issue #96, opened 2019-10-18
Invalid invocation when creating a TimeoutFuture


Guara92:
### Describe the Bug

I'm playing with the seed real world example and i got a strange exception on loading the application 

### Actual Behavior
This is the function that throw the exception:
```rust
pub fn notify_on_slow_load<Ms>(
    msg: Ms,
    error_msg: Ms,
) -> impl Future<Item = Ms, Error = Ms> {
    TimeoutFuture::new(SLOW_LOADING_THRESHOLD_MS)
        .map(|_| msg)
        .map_err(|_| error_msg)
}
```
called by the init function of the home page
```rust
pub fn init(session: Session, orders: &mut impl Orders<Msg, GMsg>) -> Model {
    let selected_feed = session
        .viewer()
        .cloned()
        .map_or_else(SelectedFeed::default, SelectedFeed::Your);

    orders
        .perform_cmd(loading::notify_on_slow_load(
            Msg::SlowLoadThresholdPassed,
            Msg::Unreachable,
        ))
        .perform_cmd(request::tag::load_list(Msg::TagsLoadCompleted))
        .perform_cmd(fetch_feed(
            session.viewer().cloned(),
            &selected_feed,
            PageNumber::default(),
        ));

    Model {
        session,
        selected_feed,
        ..Model::default()
    }
}
```

This is the exception
```
Uncaught (in promise) TypeError: Illegal invocation
    at __wbg_clearTimeout_42a8676f07d366c5 (app.e9527bfd8d64e7fbc4e9.js:89)
    at gloo_timers::sys::clear_timeout::h6b61a7affc7bb722 (:8000/wasm-function[15637]:0x604aa5)
    at <gloo_timers::future::TimeoutFuture as core::ops::drop::Drop>::drop::hd2d01337e37c0735 (:8000/wasm-function[9659]:0x56e910)
    at core::ptr::real_drop_in_place::hb3ee1e72a4dd9ef3 (:8000/wasm-function[14908]:0x5f5f75)
    at core::ptr::real_drop_in_place::h25ccf3a5381ce995 (:8000/wasm-function[14910]:0x5f601b)
    at core::ptr::real_drop_in_place::had8e8382d198844d (:8000/wasm-function[14983]:0x5f779c)
    at core::ptr::real_drop_in_place::h5feea4a304ff2612 (:8000/wasm-function[11962]:0x5b23d6)
[...]

PR #98, opened 2019-11-13
Migrate to futures 0.3

### LINKED ISSUE #97: Time to revisit gloo now async/await has landed?
Now we have async/await, is it time to revisit this library and add more stuff? :)

derekdreery:
This patch moves all crates & tests over to v0.3 (including doc tests). Closes #97 

I need to check if there is any other documentation that needs to be updated.

Also there are a few style decisions to be made as regards to async style: for example I've done

```rust
spawn_local(async {
    let mut interval = IntervalStream::new(1_000);
    loop {
        interval.next().await; // I'm discarding the result as I know it will go on for ever (might not be true if someone cancels the interval from js or using its i32 id)
        // Do something every minute
    }
})

### CODE REVIEW

**Comment by derekdreery** [comment]
closes #97

**Comment by derekdreery** [comment]
CI is failing because it is trying to use an older stable rust. Is there anything I need to do the .asure file?

**Comment by Pauan** [comment]
(Sorry for the delay on this, it's a meaty PR)

File: crates/events/README.md [hunk 1]
```diff
@@ -20,6 +20,6 @@
   <sub>Built with 🦀🕸 by <a href="https://rustwasm.github.io/">The Rust and WebAssembly Working Group</a></sub>
 </div>
 
-Using event listeners with [`web-sys`](https://crates.io/crates/web-sys) is hard! This crate provides an [`EventListener`](https://docs.rs/gloo-events/0.1.0/gloo_events/struct.EventListener.html) type which makes it easy!
+Using event listeners with [`web-sys`](https://crates.io/crates/web-sys) is hard! This crate provides an [`EventListener`](struct.EventListener.html) type which makes it easy!
```
Pauan: These changes seem wrong, since it will cause the URL to point to `https://crates.io/crates/gloo-events/struct.EventListener.html`, not `https://docs.rs/gloo-events/0.1.1/gloo_events/struct.EventListener.html` like it should.

However, I would be in favor of changing the URL to use `^0.1.0` instead, so it will point to the latest version.

crates/events/src/lib.rs
```diff
@@ -214,14 +215,17 @@ thread_local! {
...
 /// }
 ///
 /// impl Stream for OnClick {
 ///     type Item = ();
-///     type Error = ();
 ///
-///     fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
-///         self.receiver.poll().map_err(|_| unreachable!())
+///     fn poll_next(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
+///         self.project_receiver().poll_next(cx)
```
Pauan: This doesn't need `project_receiver` or unsafe code, it can just do this:

```rust
Pin::new(&mut self.receiver).poll_next(cx)
```

crates/events/tests/web.rs
```diff
@@ -38,7 +37,7 @@ where
 
 #[derive(Clone)]
 struct Sender<A> {
-    sender: mpsc::UnboundedSender<Result<A, JsValue>>,
+    sender: futures::channel::mpsc::UnboundedSender<Result<A, JsValue>>,
```
Pauan: I'd rather this continue to use the import at the top for brevity.

crates/events/tests/web.rs
```diff
@@ -50,29 +49,29 @@ impl<A> Sender<A> {
...
 // ----------------------------------------------------------------------------
 // Tests begin here
 // ----------------------------------------------------------------------------
 
-#[wasm_bindgen_test(async)]
-fn new_with_options() -> impl Future<Item = (), Error = JsValue> {
-    mpsc(|sender| {
+#[wasm_bindgen_test]
+async fn new_with_options() {
+    let results = mpsc::<(), _>(|sender| {
```
Pauan: Is this type annotation needed? If so, I would prefer if it were `let results: Vec<()> = mpsc(|sender| {` instead.

crates/events/tests/web.rs
```diff
@@ -93,13 +92,13 @@ fn new_with_options() -> impl Future<Item = (), Error = JsValue> {
 
         body.click();
         body.click();
-    })
-    .and_then(|results| is(results, vec![(), ()]))
+    }).await;
+    is(results, vec![(), ()]).unwrap_throw();
```
Pauan: This can be replaced with `assert_eq`

crates/events/tests/web.rs
```diff
@@ -50,29 +49,29 @@ impl<A> Sender<A> {
...
     F: FnOnce(Sender<A>),
 {
-    let (sender, receiver) = futures::sync::mpsc::unbounded();
+    let (sender, receiver) = futures::channel::mpsc::unbounded();
 
     f(Sender { sender });
 
     receiver
-        .then(|x| match x {
+        .map(|x| match x {
```
Pauan: This doesn't seem right, it's not handling the `Result` properly. It should be something like this instead:

```rust
receiver.try_collect().await
```

crates/timers/src/future.rs
```diff
@@ -1,8 +1,11 @@
 //! `Future`- and `Stream`-backed timers APIs.
 
+use std::task::{Poll, Context};
+use std::pin::Pin;
+use std::future::Future;
 use super::sys::*;
```
Pauan: Could you move the super import to the top? I like to keep imports roughly grouped by category.

crates/timers/src/future.rs
```diff
@@ -1,8 +1,11 @@
...
-use futures::prelude::*;
-use futures::sync::mpsc;
+use futures::stream::Stream;
```
Pauan: We should be using individual crates like `futures_core` rather than the blanket `futures`.

crates/timers/src/future.rs
```diff
@@ -24,27 +27,20 @@ use wasm_bindgen_futures::JsFuture;
...
-/// let timeout_a = TimeoutFuture::new(1_000).map(|_| "a");
-/// let timeout_b = TimeoutFuture::new(2_000).map(|_| "b");
-///
-/// wasm_bindgen_futures::spawn_local(
-///     timeout_a
-///         .select(timeout_b)
-///         .and_then(|(who, other)| {
-///             // `timeout_a` should have won this race.
-///             assert_eq!(who, "a");
+/// let timeout_a = TimeoutFuture::new(1_000);
```
Pauan: Can these be moved inside of the `async` block?

crates/timers/src/future.rs
```diff
@@ -24,27 +27,20 @@ use wasm_bindgen_futures::JsFuture;
...
-///         .map_err(|_| {
-///             wasm_bindgen::throw_str(
-///                 "unreachable -- timeouts never fail, only potentially hang"
-///             );
-///         })
-/// );
+///             drop(b);
+///         }
+///         future::Either::Right((a, val)) => {
+///             panic!("timeout_a should have won this race")
```
Pauan: Nit: `;` at the end of the line.

crates/timers/src/future.rs
```diff
@@ -24,27 +27,20 @@ use wasm_bindgen_futures::JsFuture;
...
-///         .select(timeout_b)
-///         .and_then(|(who, other)| {
-///             // `timeout_a` should have won this race.
-///             assert_eq!(who, "a");
+/// let timeout_a = TimeoutFuture::new(1_000);
+/// let timeout_b = TimeoutFuture::new(2_000);
 ///
+/// wasm_bindgen_futures::spawn_local(async {
+///     match future::select(timeout_a, timeout_b).await {
+///         future::Either::Left((val, b)) => {
```
Pauan: I'd prefer if these were all imported with `use`

crates/timers/src/future.rs
```diff
@@ -75,11 +71,10 @@ impl TimeoutFuture {
     /// use futures::prelude::*;
     /// use gloo_timers::future::TimeoutFuture;
     ///
-    /// wasm_bindgen_futures::spawn_local(
-    ///     TimeoutFuture::new(1_000).map(|_| {
-    ///         // Do stuff after one second...
-    ///     })
-    /// );
+    /// wasm_bindgen_futures::spawn_local(async {
```
Pauan: Ditto, this should be imported with `use`

crates/timers/src/future.rs
```diff
@@ -90,18 +85,25 @@ impl TimeoutFuture {
...
+    type Output = ();
 
-    fn poll(&mut self) -> Poll<(), ()> {
-        match self.inner.poll() {
-            Ok(Async::Ready(_)) => Ok(Async::Ready(())),
-            Ok(Async::NotReady) => Ok(Async::NotReady),
-            // We only ever `resolve` the promise, never reject it.
-            Err(_) => wasm_bindgen::throw_str("unreachable"),
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        match self.project_inner().poll(cx) {
```
Pauan: Similar to above, this doesn't need `unsafe` code, it can just use `Pin::new(&mut self.inner).poll(cx)`

Also, this entire thing can actually be massively simplified by using `set_timeout` directly (not using `JsFuture`). But we can save that for another PR if you like.

crates/timers/src/future.rs
```diff
@@ -136,13 +138,13 @@ impl IntervalStream {
...
-    /// wasm_bindgen_futures::spawn_local(
-    ///     IntervalStream::new(1_000)
-    ///         .for_each(|_| {
-    ///             // Do stuff every one second...
-    ///             Ok(())
-    ///         })
-    /// );
+    /// wasm_bindgen_futures::spawn_local(async {
+    ///     let mut interval = IntervalStream::new(1_000);
+    ///     loop {
```
Pauan: This should use `StreamExt::for_each`, like this:

```rust
spawn_local(async {
    IntervalStream::new(1_000).for_each(|_| {
        // Do stuff every one second...
    }).await;
});
```

crates/timers/src/future.rs
```diff
@@ -169,16 +177,15 @@ impl Drop for IntervalStream {
...
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
         if self.id.is_none() {
             self.id = Some(set_interval(
                 self.closure.as_ref().unchecked_ref::<js_sys::Function>(),
                 self.millis as i32,
             ));
         }
 
-        self.inner.poll()
+        self.project_inner().poll_next(cx)
```
Pauan: Similar to above, this can just be `Pin::new(&mut self.inner).poll_next(cx)`

crates/timers/src/lib.rs
```diff
@@ -38,14 +38,12 @@ use futures::prelude::*;
...
-let timeout = TimeoutFuture::new(1_000).and_then(|_| {
-    // Do something here after the one second timeout is up!
-#   Ok(())
-});
-
 // Spawn the `timeout` future on the local thread. If we just dropped it, then
 // the timeout would be cancelled with `clearTimeout`.
-spawn_local(timeout);
+spawn_local(async {
+    TimeoutFuture::new(1_000_).await;
```
Pauan: Nit: `_` is not needed at the end.

crates/timers/tests/web.rs
```diff
@@ -44,72 +45,62 @@ fn timeout_cancel() -> impl Future<Item = (), Error = wasm_bindgen::JsValue> {
...
+#[wasm_bindgen_test]
+async fn timeout_future_cancel() {
     let cell = Rc::new(Cell::new(false));
 
     let a = TimeoutFuture::new(1).map({
         let cell = cell.clone();
         move |_| {
             assert_eq!(cell.get(), false);
-            1
+            1 as u32
```
Pauan: Why does this need a cast? Assuming the type inference is failing, it can use `1u32` instead.

crates/events/README.md
derekdreery: Hmm, I don't remember making this change. Are the READMEs auto-generated? Anyway it's no trouble to revert it. if necessary.

crates/events/src/lib.rs
derekdreery: I feel like I need to get more of an intuition for `Pin`s. Does this work because `UnboundedReceiver<()>` is `Unpin`?

crates/timers/src/future.rs
derekdreery: If we use `set_timeout`, should the timer start when the future is created (`TimeoutFuture::new`), or when it is actually `poll`ed?

derekdreery: The current impl uses the `TimeoutFuture::new` tick to initiate the timeout. This results in a slightly more complex state machine.

crates/events/README.md
derekdreery: I've changed it to `^0.1.0` like you suggested.

derekdreery:
I've addressed all your feedback. Mostly it was just fixing things, but in the case of the future impl for timeouts I'd appreciate it if you'd have a look.

derekdreery:
I added an example just to convince myself that intervals are working. I'm happy to remove it if you prefer.

crates/events/src/lib.rs
Pauan: Yes, `Unpin` is similar to other auto-traits like `Sync` or `Send`: if all of the fields of your struct are `Unpin`, then the struct as a whole is `Unpin`. Almost all types in Rust are `Unpin`.

And if a struct is `Unpin` then you can safely access its fields (because of [this `DerefMut` impl](https://doc.rust-lang.org/nightly/std/pin/struct.Pin.html#impl-DerefMut)), and you can also safely use `Pin::new(&mut foo)` (because [`new` has an `Unpin` bound](https://doc.rust-lang.org/nightly/std/pin/struct.Pin.html#method.new)).

So the only time you need to use unsafe code is when one (or more) of your fields is *not* `Unpin`. This usually only happens in Future/Stream *combinators*, but not with leaf types like `OnClick`.

Or to put it another way: the burden of unsafe code is only for the creators of Future/Stream combinators, and not anybody else.

crates/timers/src/future.rs
Pauan: That's a really good question. I think there's reasonable arguments for both behaviors, but having it activate when polling is more logically consistent with `async`.

In any case we can discuss that in a new PR.

derekdreery:
Friendly ping :)

heilhead:
Any news on this? Excited to try out!

crates/timers/src/future.rs
```diff
@@ -71,37 +54,32 @@ impl TimeoutFuture {
...
-        match self.inner.poll() {
-            Ok(Async::Ready(_)) => Ok(Async::Ready(())),
-            Ok(Async::NotReady) => Ok(Async::NotReady),
-            // We only ever `resolve` the promise, never reject it.
-            Err(_) => wasm_bindgen::throw_str("unreachable"),
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        match Future::poll(Pin::new(&mut self.rx), cx) {
+            Poll::Pending => Poll::Pending,
+            Poll::Ready(Ok(t)) => Poll::Ready(t),
+            Poll::Ready(Err(e)) => panic!("{}", e),
```
derekdreery: Note to self - I need to put the comment back in here to explain why panicking is ok (that branch is unreachable)

crates/timers/src/future.rs
```diff
@@ -71,37 +54,32 @@ impl TimeoutFuture {
...
+    /// });
     /// ```
     pub fn new(millis: u32) -> TimeoutFuture {
-        let mut id = None;
-        let promise = js_sys::Promise::new(&mut |resolve, _reject| {
-            id = Some(set_timeout(&resolve, millis as i32));
+        let (tx, rx) = oneshot::channel();
+        let inner = Timeout::new(millis, move || {
+            // if the receiver was dropped we do nothing.
+            let _ = tx.send(());
```
Pauan: I think this should `unwrap_throw`, just as an extra precaution.

crates/timers/src/future.rs
```diff
@@ -71,37 +54,32 @@ impl TimeoutFuture {
...
+    type Output = ();
 
-    fn poll(&mut self) -> Poll<(), ()> {
-        match self.inner.poll() {
-            Ok(Async::Ready(_)) => Ok(Async::Ready(())),
-            Ok(Async::NotReady) => Ok(Async::NotReady),
-            // We only ever `resolve` the promise, never reject it.
-            Err(_) => wasm_bindgen::throw_str("unreachable"),
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        match Future::poll(Pin::new(&mut self.rx), cx) {
```
Pauan: I'd rather this be written as `Pin::new(&mut self.rx).poll(cx).map(|t| t.unwrap_throw())`

crates/timers/src/future.rs
```diff
@@ -132,53 +108,31 @@ impl IntervalStream {
...
-            millis,
-            id: None,
-            closure,
-            inner: receiver,
-        }
-    }
-}
+        let inner = Interval::new(millis, move || {
+            // if the receiver was dropped we do nothing.
+            let _ = sender.unbounded_send(());
```
Pauan: Similarly, I like `unwrap_throw` for extra safety.

crates/timers/tests/web.rs
```diff
@@ -68,48 +66,38 @@ fn timeout_future_cancel() -> impl Future<Item = (), Error = wasm_bindgen::JsVal
         let cell = cell.clone();
         move |_| {
             cell.set(true);
-            2
+            2 as u32
```
Pauan: Should be `2u32`

examples/clock/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "clock"
+version = "0.1.0"
+authors = ["Richard Dodd <richard.o.dodd@gmail.com>"]
```
Pauan: This needs to be `authors = ["Rust and WebAssembly Working Group"]`

examples/clock/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+version = "0.3.19"
+features = [
+    "console",
+    "Window",
+    "Document",
+    "Element",
+    "Node",
+]
+
+[workspace]
```
Pauan: Is there a reason for this `[workspace]`?

examples/clock/src/lib.rs
```diff
@@ -0,0 +1,30 @@
+use wasm_bindgen::prelude::*;
+use wasm_bindgen_futures::spawn_local;
+use gloo::timers::future::IntervalStream;
+use futures_util::{stream::StreamExt, future::ready};
+use chrono::Timelike;
+
+#[wasm_bindgen(start)]
+pub fn main() {
+    console_error_panic_hook::set_once();
+    let document = web_sys::window().unwrap_throw().document().unwrap_throw();
```
Pauan: Nit: could you add some empty lines in here to spread things out so they look nicer?

examples/clock/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+    spawn_local(async move {
+        IntervalStream::new(1_000).for_each(|_| {
+            render_date(&el);
+            ready(())
+        }).await;
+    });
+}
+
+fn render_date(el: &web_sys::Element) {
+    let date = chrono::Local::now();
```
Pauan: Nit: same here, a few newlines for readability.

crates/timers/tests/web.rs
derekdreery: I think the type system can infer this now. However, I can be explicit if you prefer.

crates/timers/src/future.rs
derekdreery: Yep you're right if the receiver is dropped then the whole thing is dropped so this closure should never run.

crates/events/src/lib.rs
```diff
@@ -218,10 +223,9 @@ thread_local! {
 ///
 /// impl Stream for OnClick {
 ///     type Item = ();
-///     type Error = ();
 ///
-///     fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
-///         self.receiver.poll().map_err(|_| unreachable!())
+///     fn poll_next(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
```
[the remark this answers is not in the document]
derekdreery: I wonder why my testing didn't catch this.

examples/clock/Cargo.toml
derekdreery: It's here so that the root `Cargo.toml` doesn't complain about it not being in the root workspace. (try removing it and building to see the problem).

examples/clock/src/lib.rs
derekdreery: I've tried to group things together in a sensible way, could you have a look and tell me what you think.

derekdreery:
Ready for review again.

heilhead:
I'm getting a bunch of errors trying to build now (`cargo build --target wasm32-unknown-unknown --features futures`):
```
error[E0599]: no method named `unwrap_throw` found for type `std::result::Result<(), ()>` in the current scope
  --> crates/timers/src/future.rs:69:25
   |
69 |             tx.send(()).unwrap_throw();
   |                         ^^^^^^^^^^^^ help: there is a method with a similar name: `unwrap_err`
   |
   = help: items from traits can only be used if the trait is in scope
   = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
           `use wasm_bindgen::UnwrapThrowExt;`

error[E0599]: no method named `unwrap_throw` found for type `std::result::Result<(), futures_channel::oneshot::Canceled>` in the current scope
  --> crates/timers/src/future.rs:79:60
   |
79 |         Future::poll(Pin::new(&mut self.rx), cx).map(|t| t.unwrap_throw())
   |                                                            ^^^^^^^^^^^^ help: there is a method with a similar name: `unwrap_err`
   |
   = help: items from traits can only be used if the trait is in scope
   = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
           `use wasm_bindgen::UnwrapThrowExt;`

error[E0599]: no method named `unwrap_throw` found for type `std::result::Result<(), futures_channel::mpsc::TrySendError<()>>` in the current scope
   --> crates/timers/src/future.rs:121:39
    |
121 |             sender.unbounded_send(()).unwrap_throw();
    |                                       ^^^^^^^^^^^^ help: there is a method with a similar name: `unwrap_err`
    |
    = help: items from traits can only be used if the trait is in scope
    = note: the following trait is implemented but not in scope, perhaps add a `use` for it:
            `use wasm_bindgen::UnwrapThrowExt;`

error: aborting due to 3 previous errors
```

crates/events/src/lib.rs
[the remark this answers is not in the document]
Pauan: Yes, that's odd. Maybe it's not running it? Try putting in something that's an obvious error (like an undefined variable) and see if it catches it.

examples/clock/Cargo.toml
Pauan: Ah, okay.

Pauan:
@heilhead Yeah, it needs to add in a `use wasm_bindgen::prelude::*` in order to fix that.

examples/clock/src/lib.rs
Pauan: I'd personally prefer a newline between every statement, but I guess it doesn't matter too much.

derekdreery: @Pauan I've added newlines - is this what you mean?

derekdreery:
Thanks for spotting the build problem. @Pauan is there a way I can fix the azure script so it will build this patch? The problem is that it is using rustc 1.38 which is the version before async/await. I don't know why it pulls this version in as the ci scripts say to use latest stable.

examples/clock/src/lib.rs
Pauan: Looks great, thanks!

Pauan:
@derekdreery Okay, I think I fixed the CI stuff, could you rebase and push?

Pauan:
Looks good, thanks a lot for this!

zimond:
@Pauan can you please pblish a new version of `gloo-timers` with this update?

Pauan:
@zimond Yeah, I need to fix https://github.com/rustwasm/gloo/issues/96 first before I publish.

Pauan:
@zimond Version `0.2.0` is out now!

zimond:
haha, thanks

PR #106, opened 2020-01-03
Fix timers to work in workers too.

daxpedda:
This has been fixed before in a different way by #56.
But was removed again by #103 to solve #96.

I think the solution offered in this PR should solve #96 without breaking the fix provided by #103.

Related: https://github.com/rustwasm/wasm-bindgen/issues/1046.

### DIFF
--- a/crates/timers/src/callback.rs
+++ b/crates/timers/src/callback.rs
@@ -1,25 +1,50 @@
 //! Callback-style timer APIs.
 
-use js_sys::Reflect;
+use js_sys::{Function, Reflect};
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::{throw_val, JsCast, JsValue};
 use web_sys::{Window, WorkerGlobalScope};
 
-macro_rules! global {
-    ($global:ident, $fun:expr) => {{
+thread_local! {
+    static GLOBAL: WindowOrWorker = WindowOrWorker::new();
+}
+
+enum WindowOrWorker {
+    Window(Window),
+    Worker(WorkerGlobalScope),
+}
+
+impl WindowOrWorker {
+    fn new() -> Self {
         let global: JsValue = js_sys::global().into();
 
         if Reflect::has(&global, &String::from("Window").into()).unwrap_throw() {
-            let $global: Window = global.into();
-            $fun
+            WindowOrWorker::Window(global.into())
         } else if Reflect::has(&global, &String::from("WorkerGlobalScope").into()).unwrap_throw() {
-            let $global: WorkerGlobalScope = global.into();
-            $fun
+            WindowOrWorker::Worker(global.into())
         } else {
-            throw_val(global.into())
+            throw_val(global)
         }
-    }};
+    }
 }
+
+macro_rules! impl_window_or_worker {
+    ($name:ident, ($($par_name:ident: $par_type:ty),*)$(, $return:ty)?) => {
+        impl WindowOrWorker {
+            fn $name(&self, $($par_name: $par_type),*) $( -> $return)? {
+                match self {
+                    Self::Window(window) => window.$name($($par_name),*),
+                    Self::Worker(worker) => worker.$name($($par_name),*),
+                }
+            }
+        }
+    };
+}
+
+impl_window_or_worker!(set_timeout_with_callback_and_timeout_and_arguments_0, (handler: &Function, timeout: i32), Result<i32, JsValue>);
+impl_window_or_worker!(clear_timeout_with_handle, (handle: i32));
+impl_window_or_worker!(clear_interval_with_handle, (handle: i32));
+impl_window_or_worker!(set_interval_with_callback_and_timeout_and_arguments_0, (handler: &Function, timeout: i32), Result<i32, JsValue>);
 
 /// A scheduled timeout.
 ///
@@ -37,7 +62,7 @@ pub struct Timeout {
 impl Drop for Timeout {
     fn drop(&mut self) {
         if let Some(id) = self.id {
-            global!(global, global.clear_timeout_with_handle(id));
+            GLOBAL.with(|global| global.clear_timeout_with_handle(id));
         }
     }
 }
@@ -61,15 +86,14 @@ impl Timeout {
     {
         let closure = Closure::once(callback);
 
-        let id = global!(
-            global,
+        let id = GLOBAL.with(|global| {
             global
                 .set_timeout_with_callback_and_timeout_and_arguments_0(
                     closure.as_ref().unchecked_ref::<js_sys::Function>(),
                     millis as i32,
                 )
                 .unwrap_throw()
-        );
+        });
 
         Timeout {
             id: Some(id),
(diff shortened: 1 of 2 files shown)


crates/timers/src/callback.rs
```diff
@@ -1,8 +1,25 @@
 //! Callback-style timer APIs.
 
+use js_sys::Reflect;
 use wasm_bindgen::prelude::*;
-use wasm_bindgen::JsCast;
-use web_sys::window;
+use wasm_bindgen::{throw_val, JsCast, JsValue};
+use web_sys::{Window, WorkerGlobalScope};
+
+macro_rules! global {
```
Pauan: I'd rather this be a function, it can return an enum like this:

```rust
enum WindowOrWorker {
    Window(Window),
    Worker(WorkerGlobalScope),
}
```

Then it could have a macro which would generate methods like this:

```rust
impl WorkerOrGlobal {
    fn foo(&self) {
        match self {
            Self::Window(window) => window.foo(),
            Self::Worker(worker) => worker.foo(),
        }
    }
}
```

crates/timers/src/callback.rs
```diff
@@ -1,8 +1,25 @@
...
+    ($global:ident, $fun:expr) => {{
+        let global: JsValue = js_sys::global().into();
+
+        if Reflect::has(&global, &String::from("Window").into()).unwrap_throw() {
```
Pauan: I think this would be better written like this:

```rust
match js_sys::global().dyn_into::<Window>() {
    Ok(window) => WindowOrGlobal::Window(window),
    Err(global) => {
        let worker = global.dyn_into::<WorkerGlobalScope>().unwrap_throw();
        WindowOrGlobal::Worker(worker)
    },
}
```

crates/timers/src/callback.rs
Pauan: Oh, also, it should cache the `WindowOrWorker` inside a `thread_local!`, so that way it only needs to generate it once.

crates/timers/src/callback.rs
daxpedda: This doesn't work, `dyn_into` will use `typeof Window` inside of the generated javascript, which will throw an error saying that `Window` doesn't exist.

crates/timers/src/callback.rs
daxpedda: Done, is that implementation satisfactory?

crates/timers/src/callback.rs
Pauan: Hmmm, that's a really tricky problem. Let me think about that.

crates/timers/src/callback.rs
```diff
@@ -1,8 +1,50 @@
...
+        } else if Reflect::has(&global, &String::from("WorkerGlobalScope").into()).unwrap_throw() {
+            WindowOrWorker::Worker(global.into())
+        } else {
+            throw_val(global)
+        }
+    }
+}
+
+macro_rules! impl_window_or_worker {
+    ($name:ident, ($($par_name:ident: $par_type:ty),*)$(, $return:ty)?) => {
```
Pauan: I think this would be better like this:

```rust
macro_rules! impl_window_or_worker {
    ($(fn $name:ident($($par_name:ident: $par_type:ty),*)$( -> $return:ty)?;)+) => {
        impl WindowOrWorker {
            $(
                fn $name(&self, $($par_name: $par_type),*)$( -> $return)? {
                    match self {
                        Self::Window(window) => window.$name($($par_name),*),
                        Self::Worker(worker) => worker.$name($($par_name),*),
                    }
                }
            )+
        }
    };
}

impl_window_or_worker! {
    fn set_timeout_with_callback_and_timeout_and_arguments_0(handler: &Function, timeout: i32) -> Result<i32, JsValue>;
    fn set_interval_with_callback_and_timeout_and_arguments_0(handler: &Function, timeout: i32) -> Result<i32, JsValue>;
    fn clear_timeout_with_handle(handle: i32);
    fn clear_interval_with_handle(handle: i32);
}
```

daxpedda: Sure, I just looked over it and copy pasted.

crates/timers/src/callback.rs
Pauan: Okay, so I think we should use this solution as a temporary measure until we get better support in wasm-bindgen:

```rust
impl WindowOrWorker {
    fn new() -> Self {
        #[wasm_bindgen]
        extern "C" {
            type Global;

            #[wasm_bindgen(method, getter, js_name = Window)]
            fn window(this: &Global) -> JsValue;

            #[wasm_bindgen(method, getter, js_name = WorkerGlobalScope)]
            fn worker(this: &Global) -> JsValue;
        }

        let global: Global = js_sys::global().unchecked_into();

        if !global.window().is_undefined() {
            Self::Window(global.unchecked_into())

        } else if !global.worker().is_undefined() {
            Self::Worker(global.unchecked_into())

        } else {
            panic!("Only supported in a browser or web worker");
        }
    }
}
```

daxpedda: May I ask why? What's wrong with `Reflect`?

Pauan: It's slow, because it's more dynamic, it has to do more checks, and it has to serialize strings to JS. This way is faster.

daxpedda: May I ask why? What's wrong with `Reflect`?

Pauan: It's slow, because it's more dynamic, it has to do more checks, and it has to serialize strings to JS. This way is faster.

daxpedda: Aha! Thanks for the expanation!

Pauan: The extra checks involve both a `try`/`catch` in JS *and* some extra checks on the Rust side, you can see the performance difference clearly by looking at the generated JS code:

```js
export const __wbg_Window_28cae7b222075c59 = function(arg0) {
    const ret = getObject(arg0).Window;
    return addHeapObject(ret);
};
```

vs

```js
let cachedTextDecoder = new TextDecoder('utf-8');

function getStringFromWasm(ptr, len) {
    return cachedTextDecoder.decode(getUint8Memory().subarray(ptr, ptr + len));
}

function getCachedStringFromWasm(ptr, len) {
    if (ptr === 0) {
        return getObject(len);
    } else {
        return getStringFromWasm(ptr, len);
    }
}

function handleError(e) {
    wasm.__wbindgen_exn_store(addHeapObject(e));
}

export const __wbg_get_003e1b80a63de7c5 = function(arg0, arg1) {
    try {
        const ret = Reflect.get(getObject(arg0), getObject(arg1));
        return addHeapObject(ret);
    } catch (e) {
        handleError(e)
    }
};
```

daxpedda: Done, again I just double-checked and copy pasted.
Yeah I was looking at it right now, I find this very strange, but I didn't delve enough into `wasm_bindgen` to really know my way around all of this, anyways, thanks for helping out and for the explanation!

Pauan: As a general rule, if you know the properties ahead of time you should use static bindings with `extern "C" { ... }`.

So you would only use `Reflect` when you need to look up things dynamically, like `Reflect::get(&foo, &bar)`

The way to think of it is that `extern "C" { ... }` corresponds to `foo.bar` in JS, whereas `Reflect` corresponds to `foo[bar]` in JS.

daxpedda: I see, thanks!

derekdreery: This is good advice and should live in the wasm-bindgen book somewhere I think.

PR #107, opened 2020-01-07
requestAnimationFrame RFC

dakom:
initial draft of the rfc for requestAnimationFrame API

### DIFF
--- a/rfcs/002-mid-level-raf-api.md
+++ b/rfcs/002-mid-level-raf-api.md
@@ -1,0 +1,72 @@
+# Mid-Level requestAnimationFrame Api
+
+## Summary
+
+This is a mid-level API wrapper around the requestAnimationFrame JavaScript API. The possibility for a higher level API is left open.
+
+## The API
+
+The public API is a `Raf` struct, created with `FnMut(f64)` as a parameter to `Raf::new()`. 
+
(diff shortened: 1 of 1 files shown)


dakom:
Should I add `RafState` into the PR?

dakom:
added it in :)

rfcs/002-mid-level-raf-api.md
```diff
@@ -0,0 +1,77 @@
...
+    }
+}
+
+```
+
+## Prior Art
+
+The initial commit of this RFC uses the code verbatim from [dominator](https://github.com/Pauan/rust-dominator/) (with permission from the author)
+
+The wasm-bindgen guide proposes [another approach](https://rustwasm.github.io/wasm-bindgen/examples/request-animation-frame.html) which is similar in spirit but less gloo-like. Specifically, adapting this into a generic API that supports cancellation would likely require that the user call a `cancel()` function explicitly, rather than the RAII style proposed here.
```
yoshuawuyts: I'm unsure about the meaning of "gloo-like" here; I don't feel like we quite have concensus on what "gloo" APIs should be like, so perhaps leave this part out? The next sentence feels clear in explaining what the differences are.

dakom: good catch - just pushed a change with that phrase removed

PR #110, opened 2020-01-18
Files implementation

### LINKED ISSUE #47: File API Library
Let's make an idiomatic, Rust-y interface to the file API!

https://w3c.github.io/FileAPI/

derekdreery:
This carries on from the work in #83 from @rylev. I've tried to address the issues in that thread.

Closes #47

### DIFF
--- a/crates/file/src/blob.rs
+++ b/crates/file/src/blob.rs
@@ -1,0 +1,344 @@
+use std::{
+    ops::Deref,
+    time::{Duration, SystemTime, UNIX_EPOCH},
+};
+
+use wasm_bindgen::{prelude::*, throw_str, JsCast};
+
+mod sealed {
+    pub trait Sealed {}
+}
+use sealed::Sealed;
+
+/// This trait is used to overload the `Blob::new_with_options` function, allowing a variety of
+/// types to be used to create a `Blob`. Ignore this, and use &[u8], &str, etc to create a `Blob`.
+///
+/// The trait is sealed: it can only be implemented by types in this
+/// crate, as this crate relies on invariants regarding the `JsValue` returned from `into_jsvalue`.
+pub trait BlobContents: Sealed {
+    fn into_jsvalue(self) -> JsValue;
+}
+
+impl<'a> Sealed for &'a str {}
+impl<'a> BlobContents for &'a str {
+    fn into_jsvalue(self) -> JsValue {
+        JsValue::from_str(self)
+    }
+}
+
+impl<'a> Sealed for &'a [u8] {}
+impl<'a> BlobContents for &'a [u8] {
+    fn into_jsvalue(self) -> JsValue {
+        js_sys::Uint8Array::from(self).into()
+    }
+}
+
+impl Sealed for js_sys::ArrayBuffer {}
+impl BlobContents for js_sys::ArrayBuffer {
+    fn into_jsvalue(self) -> JsValue {
+        self.into()
+    }
+}
+
+/// A [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob).
+///
+/// `Blob`s can be created directly from `&str`, `&[u8]`, and `js_sys::ArrayBuffer`s using the
+/// `Blob::new` or `Blob::new_with_options` functions.
+#[derive(Debug)]
+pub struct Blob {
+    inner: web_sys::Blob,
+}
+
+impl Blob {
+    /// Create a new `Blob` from a `&str`, `&[u8]` or `js_sys::ArrayBuffer`.
+    pub fn new<T>(content: T) -> Blob
+    where
+        T: BlobContents,
+    {
+        Blob::new_with_options(content, None)
+    }
+
+    /// Like `new`, but allows specifying the MIME type (also known as *content type* or *media
+    /// type*) of the `Blob`.
+    pub fn new_with_options<T>(content: T, mime_type: Option<&str>) -> Blob
+    where
+        T: BlobContents,
+    {
+        let mut properties = web_sys::BlobPropertyBag::new();
+        if let Some(mime_type) = mime_type {
+            properties.type_(mime_type);
+        }
+
+        let parts = js_sys::Array::of1(&content.into_jsvalue());
+        let inner = web_sys::Blob::new_with_u8_array_sequence_and_options(&parts, &properties);
+
+        Blob::from(inner.unwrap_throw())
+    }
+
+    pub fn slice(&self, start: u64, end: u64) -> Self {
+        let start = safe_u64_to_f64(start);
+        let end = safe_u64_to_f64(end);
+
+        let b: &web_sys::Blob = self.as_ref();
+        Blob::from(b.slice_with_f64_and_f64(start, end).unwrap_throw())
+    }
+
+    /// The number of bytes in the Blob/File.
+    pub fn size(&self) -> u64 {
+        safe_f64_to_u64(self.inner.size())
+    }
+
+    /// The statically typed MIME type (also known as *content type* or *media type*) of the `File`
+    /// or `Blob`.
+    #[cfg(feature = "mime")]
+    pub fn mime_type(&self) -> Result<mime::Mime, mime::FromStrError> {
+        self.raw_mime_type().parse()
+    }
+
+    /// The raw MIME type (also known as *content type* or *media type*) of the `File` or
+    /// `Blob`.
+    pub fn raw_mime_type(&self) -> String {
+        self.inner.type_()
+    }
+}
+
+impl From<web_sys::Blob> for Blob {
+    fn from(blob: web_sys::Blob) -> Self {
+        Blob { inner: blob }
+    }
+}
+
+impl From<web_sys::File> for Blob {
+    fn from(file: web_sys::File) -> Self {
+        Blob { inner: file.into() }
+    }
+}
+
+impl From<Blob> for web_sys::Blob {
+    fn from(blob: Blob) -> Self {
+        blob.inner
+    }
+}
+
+impl From<Blob> for JsValue {
+    fn from(blob: Blob) -> Self {
+        blob.inner.into()
+    }
+}
+
+impl AsRef<web_sys::Blob> for Blob {
+    fn as_ref(&self) -> &web_sys::Blob {
+        self.inner.as_ref()
+    }
+}
+
+impl AsRef<JsValue> for Blob {
+    fn as_ref(&self) -> &JsValue {
+        self.inner.as_ref()
+    }
+}
+
+/// A [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File).
+#[derive(Debug)]
+pub struct File {
+    // the trick here is that we know the contents of `inner` are a file, even though that type
+    // information is not stored. It is the same trick as is used in `web_sys`.
+    inner: Blob,
+}
+
+impl File {
+    /// Create a new `File` with the given name and contents.
+    ///
+    /// `contents` can be `&str`, `&[u8]`, or `js_sys::ArrayBuffer`.
+    pub fn new<T>(name: &str, contents: T) -> File
+    where
+        T: BlobContents,
+    {
+        Self::new_(name, contents.into_jsvalue(), None, None)
+    }
+
+    /// Like `File::new`, but allows customizing the MIME type (also
+    /// known as *content type* or *media type*), and the last modified time.
+    ///
+    /// `std::time::SystemTime` is a low level type, use a crate like
+    /// [`chrono`](https://docs.rs/chrono/0.4.10/chrono/) to work with a more user-friendly
+    /// representation of time.
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// use chrono::prelude::*;
+    /// use gloo_file::File;
+    ///
+    /// // Just create a dummy `gloo::file::File` for demonstration purposes.
+    /// let example_file = File::new_with_options(
+    ///     "motivation.txt",
+    ///     "live your best life",
+    ///     Some("text/plain"),
+    ///     Some(Utc::now().into())
+    /// );
+    /// assert_eq!(example_file.name(), String::from("motivation.txt"));
+    /// assert_eq!(example_file.raw_mime_type(), String::from("text/plain"));
+    /// ```
+    pub fn new_with_options<T>(
+        name: &str,
+        contents: T,
+        mime_type: Option<&str>,
+        last_modified_time: Option<SystemTime>,
+    ) -> File
+    where
+        T: BlobContents,
+    {
+        Self::new_(name, contents.into_jsvalue(), mime_type, last_modified_time)
+    }
+
+    /// Private constructor.
+    fn new_(
+        name: &str,
+        contents: JsValue,
+        mime_type: Option<&str>,
+        last_modified_time: Option<SystemTime>,
+    ) -> Self {
+        let mut options = web_sys::FilePropertyBag::new();
+        if let Some(mime_type) = mime_type {
+            options.type_(&mime_type);
+        }
+
+        if let Some(last_modified_time) = last_modified_time {
+            let duration = match last_modified_time.duration_since(UNIX_EPOCH) {
+                Ok(duration) => safe_u128_to_f64(duration.as_millis()),
+                Err(time_err) => -safe_u128_to_f64(time_err.duration().as_millis()),
+            };
+            options.last_modified(duration);
+        }
+
+        let parts = js_sys::Array::of1(&contents);
+        let inner = web_sys::File::new_with_u8_array_sequence_and_options(&parts, &name, &options)
+            .unwrap_throw();
+
+        File::from(inner)
+    }
+
+    /// Gets the file name.
+    pub fn name(&self) -> String {
+        let f: &web_sys::File = self.as_ref();
+        f.name()
+    }
+
+    /// Gets the time that the file was last modified.
+    ///
+    /// `std::time::SystemTime` is a low level type, use a crate like
+    /// [`chrono`](https://docs.rs/chrono/0.4.10/chrono/) to work with more user-friendly
+    /// representations of time. For example:
+    ///
+    /// ```rust,no_run
+    /// use chrono::prelude::*;
+    /// use gloo_file::File;
+    ///
+    /// // Just create a dummy `gloo::file::File` for demonstration purposes.
+    /// let example_file = File::new("test_file.txt", "<almost empty contents>");
+    /// let date: DateTime<Utc> = example_file.last_modified_time().into();
+    /// ```
+    pub fn last_modified_time(&self) -> SystemTime {
+        let f: &web_sys::File = self.as_ref();
+        match f.last_modified() {
+            pos if pos >= 0.0 => UNIX_EPOCH + Duration::from_millis(safe_f64_to_u64(pos)),
+            neg => UNIX_EPOCH - Duration::from_millis(safe_f64_to_u64(-neg)),
+        }
+    }
+
+    /// Create a new `File` from a sub-part of this `File`.
+    pub fn slice(&self, start: u64, end: u64) -> Self {
+        let blob = self.deref().slice(start, end);
+
+        let raw_mime_type = self.raw_mime_type();
+        let mime_type = if raw_mime_type == "" {
+            None
+        } else {
+            Some(raw_mime_type)
+        };
+
+        File::new_(
+            &self.name(),
+            blob.into(),
+            mime_type.as_ref().map(|s| s.as_str()),
+            Some(self.last_modified_time()),
+        )
+    }
+}
+
+impl From<web_sys::File> for File {
+    fn from(file: web_sys::File) -> Self {
+        File {
+            inner: Blob::from(web_sys::Blob::from(file)),
+        }
+    }
+}
+
+impl Deref for File {
+    type Target = Blob;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+impl AsRef<web_sys::File> for File {
+    fn as_ref(&self) -> &web_sys::File {
+        <Blob as AsRef<web_sys::Blob>>::as_ref(&self.inner).unchecked_ref()
+    }
+}
+
+impl AsRef<web_sys::Blob> for File {
+    fn as_ref(&self) -> &web_sys::Blob {
+        self.inner.as_ref()
+    }
+}
+
+impl From<File> for Blob {
+    fn from(file: File) -> Self {
+        file.inner
+    }
+}
+
+// utility methods
+// ===============
+
+/// JavaScript only has `f64`, which has a maximum accurate integer size of`2^53 - 1`. So we use
+/// this to safely convert from larger integers to `f64`.  See
+/// [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)
+fn safe_u64_to_f64(number: u64) -> f64 {
+    // Max integer stably representable by f64
+    // todo use js_sys::Number::MAX_SAFE_INTEGER once stable
+    const MAX_SAFE_INTEGER: u64 = 9007199254740991; // (2^53 - 1)
+    if number > MAX_SAFE_INTEGER {
+        throw_str("a rust number was too large and could not be represented in JavaScript");
+    }
+    number as f64
+}
+
+fn safe_u128_to_f64(number: u128) -> f64 {
+    // Max integer stably representable by f64
+    // todo use js_sys::Number::MAX_SAFE_INTEGER once stable
+    const MAX_SAFE_INTEGER: u128 = 9007199254740991; // (2^53 - 1)
+    if number > MAX_SAFE_INTEGER {
+        throw_str("a rust number was too large and could not be represented in JavaScript");
+    }
+    number as f64
+}
+
+/// Like safe_u64_to_f64, but additionally checks that the number is an integer.
+fn safe_f64_to_u64(number: f64) -> u64 {
+    // Max integer stably representable by f64
+    // todo use js_sys::Number::MAX_SAFE_INTEGER once stable
+    const MAX_SAFE_INTEGER: f64 = 9007199254740991.0; // (2^53 - 1)
+    if number > MAX_SAFE_INTEGER {
+        throw_str("a rust number was too large and could not be represented in JavaScript");
+    }
+    if number.floor() != number {
+        throw_str(
+            "a number could not be converted to an integer because it was not a whole number",
+        );
+    }
+    number as u64
+}
--- a/crates/file/src/file_reader.rs
+++ b/crates/file/src/file_reader.rs
@@ -1,0 +1,272 @@
+use wasm_bindgen::throw_str;
+
+pub mod callbacks {
+    use crate::{
+        blob::Blob,
+        file_reader::{FileReadError, ReadyState},
+    };
+    use gloo_events::EventListener;
+    use std::{cell::RefCell, rc::Rc};
+    use wasm_bindgen::{prelude::*, throw_str, JsCast, UnwrapThrowExt};
+
+    /// A guard object that aborts the file read when dropped (if the read isn't already
+    /// finished).
+    #[derive(Debug)]
+    pub struct FileReader {
+        reader: web_sys::FileReader,
+        load_listener: EventListener,
+        error_listener: EventListener,
+    }
+
+    impl std::ops::Drop for FileReader {
+        fn drop(&mut self) {
+            if !ReadyState::from(self.reader.ready_state()).is_done() {
+                self.reader.abort();
+            }
+        }
+    }
+
+    /// Asynchronously converts `blob` into a text string and then passes it to the `callback`.
+    ///
+    /// If the returned `FileReader` is dropped before the callback is called, the read will be
+    /// cancelled.
+    pub fn read_as_text<F>(blob: &Blob, callback: F) -> FileReader
+    where
+        F: FnOnce(Result<String, FileReadError>) + 'static,
+    {
+        read(
+            blob,
+            callback,
+            |value| value.as_string().unwrap_throw(),
+            |reader, blob| reader.read_as_text(blob).unwrap_throw(),
+        )
+    }
+
+    /// Asynchronously converts the `blob` into a base64 encoded `data:` URL and then passes it to
+    /// the `callback`.
+    ///
+    /// If the returned `FileReader` is dropped before the callback is called, the read will be
+    /// cancelled.
+    pub fn read_as_data_url<F>(blob: &Blob, callback: F) -> FileReader
+    where
+        F: FnOnce(Result<String, FileReadError>) + 'static,
+    {
+        read(
+            blob,
+            callback,
+            |value| value.as_string().unwrap_throw(),
+            |reader, blob| reader.read_as_data_url(blob).unwrap_throw(),
+        )
+    }
+
+    /// Asynchronously converts the `blob` into an array buffer and then passes it to the `callback`.
+    ///
+    /// If the returned `FileReader` is dropped before the callback is called, the read will be
+    /// cancelled.
+    pub fn read_as_array_buffer<F>(blob: &Blob, callback: F) -> FileReader
+    where
+        F: FnOnce(Result<js_sys::ArrayBuffer, FileReadError>) + 'static,
+    {
+        read(
+            blob,
+            callback,
+            |value| value.dyn_into::<js_sys::ArrayBuffer>().unwrap_throw(),
+            |reader, blob| reader.read_as_array_buffer(blob).unwrap_throw(),
+        )
+    }
+
+    /// Asynchronously converts the `blob` into a `Vec<u8>` and then passes it to the `callback`.
+    ///
+    /// If the returned `FileReader` is dropped before the callback is called, the read will be
+    /// cancelled.
+    pub fn read_as_bytes<F>(blob: &Blob, callback: F) -> FileReader
+    where
+        F: FnOnce(Result<Vec<u8>, FileReadError>) + 'static,
+    {
+        read_as_array_buffer(blob, move |result| {
+            callback(result.map(|buffer| js_sys::Uint8Array::new(&buffer).to_vec()));
+        })
+    }
+
+    /// Generic function to start the async read of the `FileReader`.
+    ///
+    /// `callback` is the user-supplied callback, `extract_fn` function converts between JsValue
+    /// returned by the read and the type the callback expects, and `read_fn` is the method that
+    /// runs the async read on the `FileReader`.
+    fn read<T, CF, EF, RF>(blob: &Blob, callback: CF, extract_fn: EF, read_fn: RF) -> FileReader
+    where
+        CF: FnOnce(Result<T, FileReadError>) + 'static,
+        EF: Fn(JsValue) -> T + 'static,
+        RF: Fn(&web_sys::FileReader, &web_sys::Blob) + 'static,
+    {
+        // we need to be able to run the FnOnce, while proving to the compiler that it can only run
+        // once. The easiest way is to `take` it out of an Option. The `Rc` and `RefCell` are
+        // because we need shared ownership and mutability (for the `take`).
+        let load_callback: Rc<RefCell<Option<CF>>> = Rc::new(RefCell::new(Some(callback)));
+        let error_callback = load_callback.clone();
+
+        let reader = web_sys::FileReader::new().unwrap_throw();
+
+        let load_reader = reader.clone();
+        let error_reader = reader.clone();
+        // Either the load listener or the error listener will be called, never both (so FnOnce is
+        // ok).
+        let load_listener = EventListener::new(&reader, "load", move |_event| {
+            let result = extract_fn(load_reader.result().unwrap_throw());
+            let callback = load_callback.borrow_mut().take().unwrap_throw();
+            callback(Ok(result));
+        });
+
+        let error_listener = EventListener::new(&reader, "error", move |_event| {
+            let exception = error_reader.error().unwrap_throw();
+            let error = match exception.name().as_str() {
+                "NotFoundError" => FileReadError::NotFound(exception.message()),
+                "NotReadableError" => FileReadError::NotReadable(exception.message()),
+                "SecurityError" => FileReadError::Security(exception.message()),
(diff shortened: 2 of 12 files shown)


derekdreery:
Concerns: 

 - Could do with more test coverage
 - Callback interface for reading blobs could use `gloo::events`.
 - The error handling might be broken (I'm not sure how to trigger the different kinds of errors in tests).

We could address these in followup work.

derekdreery:
I've added a patch that uses a slightly different approach to the Blob -> File inheritance hierarchy, using the `Deref` trick used in `web_sys`. See #02c675c for the original implementation.

derekdreery:
There is currently a problem - there is no way to get the error that occurred because the `error` attribute is not in web-sys. I'm investigating in wasm-bindgen.

derekdreery:
Just to show how act `gloo` can be: here is how to read in a file and convert it into a data url:

```rust
async fn data_url_future() {
    let blob = Blob::new_with_options(PNG_FILE, Some("image/png"));
    assert_eq!(
        gloo_file::futures::read_as_data_url(&blob).await.unwrap(),
        PNG_FILE_DATA
    );
}

const PNG_FILE: &'static [u8] = &[
    0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a, 0x00, 0x00, 0x00, 0x0d, 0x49, 0x48, 0x44, 0x52,
    0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x01, 0x01, 0x03, 0x00, 0x00, 0x00, 0x25, 0xdb, 0x56,
    0xca, 0x00, 0x00, 0x00, 0x03, 0x50, 0x4c, 0x54, 0x45, 0xff, 0x4d, 0x00, 0x5c, 0x35, 0x38, 0x7f,
    0x00, 0x00, 0x00, 0x01, 0x74, 0x52, 0x4e, 0x53, 0xcc, 0xd2, 0x34, 0x56, 0xfd, 0x00, 0x00, 0x00,
    0x0a, 0x49, 0x44, 0x41, 0x54, 0x78, 0x9c, 0x63, 0x62, 0x00, 0x00, 0x00, 0x06, 0x00, 0x03, 0x36,
    0x37, 0x7c, 0xa8, 0x00, 0x00, 0x00, 0x00, 0x49, 0x45, 0x4e, 0x44, 0xae, 0x42, 0x60, 0x82,
];

const PNG_FILE_DATA: &'static str =
    "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABAQMAAA\
     Al21bKAAAAA1BMVEX/TQBcNTh/AAAAAXRSTlPM0jRW/QAAAApJREFUeJxjYgAAAAYAAzY3fKgAAAAASUVORK5CYII=";
```

Now imagine that you got the image from `input[type=file].files[0]`. Assuming you are using async, you can grab an image the user has selected and display it on the page in less than 5 lines of code.

derekdreery:
ping

Pauan:
Sorry for the delay on this, I'll get to it soon!

crates/file/src/lib.rs
```diff
@@ -0,0 +1,20 @@
+//! Provides wrappers for working with `Blob`s and `File`s from javascript.
```
Pauan:
```suggestion
//! Provides wrappers for working with `Blob`s and `File`s from JavaScript.
```

crates/file/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+//!
+//! A `File` is just a `Blob` with some extra data: a name and a last modified time.
+//!
+//! In the File API, `Blob`s are opaque objects that lazily fetch their contained data when
+//! asked. This allows a `Blob` to represent some resource that isn't completely available, for
+//! example a webassembly message that is being received or a file that needs to be read from disk.
```
Pauan:
```suggestion
//! example a WebSocket message that is being received or a file that needs to be read from disk.
```

crates/file/src/lib.rs
```diff
@@ -0,0 +1,20 @@
...
+//! Access to the contents of `Blob`s is gained through callbacks, allowing the user agent to defer
```
Pauan:
```suggestion
//! You can asynchronously access the contents of the `Blob` through callbacks.
//! But that is rather inconvenient, so this crate provides some functions which
//! return a `Future` instead.
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// encoded data url.
+    pub async fn read_as_data_url(blob: &Blob) -> Result<String, FileReadError> {
+        let (sender, receiver) = futures_channel::oneshot::channel();
+        let _reader = super::callbacks::read_as_data_url(blob, |result| {
+            sender.send(result).unwrap_throw();
+        });
+        receiver.await.unwrap_throw()
+    }
+
+    /// Read the contents of `blob` into an array buffer.
```
Pauan:
```suggestion
    /// Returns the contents of `blob` as an array buffer.
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+        }
+    }
+}
+
+#[cfg(feature = "futures")]
+pub mod futures {
+    use crate::{Blob, FileReadError};
+    use wasm_bindgen::UnwrapThrowExt;
+
+    /// Read the contents of `blob` into a text string.
```
Pauan:
```suggestion
    /// Returns the contents of `blob` as a text string.
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub async fn read_as_text(blob: &Blob) -> Result<String, FileReadError> {
+        let (sender, receiver) = futures_channel::oneshot::channel();
+        let _reader = super::callbacks::read_as_text(blob, |result| {
+            sender.send(result).unwrap_throw();
+        });
+        receiver.await.unwrap_throw()
+    }
+
+    /// Read the contents of `blob` into a string with the blob contents serialized as a base64
```
Pauan:
```suggestion
    /// Returns the contents of `blob` as a base64 encoded `data:` URL.
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// Read the contents of `blob` into an array buffer.
+    pub async fn read_as_array_buffer(blob: &Blob) -> Result<js_sys::ArrayBuffer, FileReadError> {
+        let (sender, receiver) = futures_channel::oneshot::channel();
+        let _reader = super::callbacks::read_as_array_buffer(blob, |result| {
+            sender.send(result).unwrap_throw();
+        });
+        receiver.await.unwrap_throw()
+    }
+
+    /// Read the contents of `blob` into a Vec<u8>.
```
Pauan:
```suggestion
    /// Returns the contents of `blob` as a `Vec<u8>`.
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    impl std::ops::Drop for FileReader {
+        fn drop(&mut self) {
+            if !ReadyState::from(self.reader.ready_state()).is_done() {
+                self.reader.abort();
+            }
+        }
+    }
+
+    /// Read the contents of `blob` into a text string, and then run the callback.
```
Pauan:
```suggestion
    /// Asynchronously converts `blob` into a text string and then passes it to the `callback`.
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+    ///
+    /// If the returned `FileReader` is dropped before the callback is invoked, the read will be
+    /// cancelled.
+    pub fn read_as_text<F>(blob: &Blob, callback: F) -> FileReader
+    where
+        F: FnOnce(Result<String, FileReadError>) + 'static,
+    {
+        // we need to be able to run the FnOnce, while proving to the compiler that it can only run
+        // once. The easiest way is to `take` it out of an Option. The `Arc` and `Mutex` are
```
Pauan: These should be `Rc` and `RefCell` instead. Using `Arc` + `Mutex` just causes an extra performance cost for no benefit.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+            let result = reader.result().unwrap_throw().as_string().unwrap_throw();
+            load_callback.lock().unwrap_throw().take().unwrap_throw()(Ok(result));
+        });
+
+        let error_listener = EventListener::new(&reader, "error", move |event| {
+            let error = error_from_event(&event);
+            error_callback.lock().unwrap_throw().take().unwrap_throw()(Err(error));
+        });
+
+        reader.read_as_text(blob.as_ref()).unwrap_throw();
```
Pauan: Nit: there should be a newline here, just for visual spacing.

```suggestion
        reader.read_as_text(blob.as_ref()).unwrap_throw();

```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// cancelled.
+    pub fn read_as_text<F>(blob: &Blob, callback: F) -> FileReader
+    where
+        F: FnOnce(Result<String, FileReadError>) + 'static,
+    {
+        // we need to be able to run the FnOnce, while proving to the compiler that it can only run
+        // once. The easiest way is to `take` it out of an Option. The `Arc` and `Mutex` are
+        // because we need shared ownership and mutability (for the `take`).
+        let load_callback: Arc<
+            Mutex<Option<Box<dyn FnOnce(Result<String, FileReadError>) + 'static>>>,
```
Pauan: Does this need `Box`, or can it get away with storing it directly, like this?

```rust
let load_callback: Rc<RefCell<Option<F>>> = Rc::new(RefCell::new(Some(callback)));
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+        > = Arc::new(Mutex::new(Some(Box::new(callback))));
+        let error_callback = load_callback.clone();
+
+        let reader = web_sys::FileReader::new().unwrap_throw();
+
+        // Either the load listener or the error listener will be called, never both (so FnOnce is
+        // ok).
+        let load_listener = EventListener::new(&reader, "load", move |event| {
+            let reader = reader_from_event(&event);
```
Pauan: I would rather this just `clone` the `reader` and pass it into the callback.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+            let result = reader.result().unwrap_throw().as_string().unwrap_throw();
+            load_callback.lock().unwrap_throw().take().unwrap_throw()(Ok(result));
+        });
+
+        let error_listener = EventListener::new(&reader, "error", move |event| {
+            let error = error_from_event(&event);
```
Pauan: Similarly, I'd rather this just `clone` the `reader`.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        reader.read_as_text(blob.as_ref()).unwrap_throw();
+        FileReader {
+            reader,
+            load_listener,
+            error_listener,
+        }
+    }
+
+    /// Read the contents of `blob` into a string with the blob contents serialized as a base64
```
Pauan:
```suggestion
    /// Asynchronously converts the `blob` into a base64 encoded `data:` URL and then passes it to the `callback`.
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// encoded data url, and then run the callback.
+    ///
+    /// If the returned `FileReader` is dropped before the callback is invoked, the read will be
+    /// cancelled.
+    pub fn read_as_data_url<F>(blob: &Blob, callback: F) -> FileReader
+    where
+        F: FnOnce(Result<String, FileReadError>) + 'static,
+    {
+        let load_callback: Arc<
```
Pauan: There's a bunch of code duplication with these, I think it'd be great if the actual loading could be factored out into a separate function.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        reader.read_as_data_url(blob.as_ref()).unwrap_throw();
+        FileReader {
+            reader,
+            load_listener,
+            error_listener,
+        }
+    }
+
+    /// Read the contents of `blob` into an array buffer, and then run the callback.
```
Pauan:
```suggestion
    /// Asynchronously converts the `blob` into an array buffer and then passes it to the `callback`.
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        reader.read_as_array_buffer(blob.as_ref()).unwrap_throw();
+        FileReader {
+            reader,
+            load_listener,
+            error_listener,
+        }
+    }
+
+    /// Read the contents of `blob` into a `Vec<u8>`, and then run the callback.
```
Pauan:
```suggestion
    /// Asynchronously converts the `blob` into a `Vec<u8>` and then passes it to the `callback`.
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub fn read_as_bytes<F>(blob: &Blob, callback: F) -> FileReader
+    where
+        F: FnOnce(Result<Vec<u8>, FileReadError>) + 'static,
+    {
+        read_as_array_buffer(blob, move |result| {
+            callback(result.map(|buffer| js_sys::Uint8Array::new(&buffer).to_vec()));
+        })
+    }
+
+    /// Get the event from the reader target, doing necessary casting.
```
Pauan: By cloning the `reader` this function won't be needed anymore.

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,268 @@
...
+        target.unchecked_into()
+    }
+
+    fn error_from_event(event: &web_sys::Event) -> FileReadError {
+        let exception = reader_from_event(event).error().unwrap_throw();
+        match exception.name().as_str() {
+            "NotFoundError" => FileReadError::NotFound(exception.message()),
+            "NotReadableError" => FileReadError::NotReadable(exception.message()),
+            "SecurityError" => FileReadError::Security(exception.message()),
+            name => throw_str(&format!("unrecognised error type: {}", name)),
```
Pauan: I'd really like to avoid the formatting code if possible, but I'm not sure what way would be best.

crates/file/src/file_reader.rs
Pauan: I really like the way you implemented these with `async fn`, very clean!

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+
+mod sealed {
+    pub trait Sealed {}
+}
+use sealed::Sealed;
+
+/// This trait is used to overload the `Blob::new_with_options` function, allowing a variety of
+/// types to be used to create a `Blob`. Ignore this, and use &[u8], &str, etc to create a `Blob`.
+///
+/// The trait is sealed: it must only be implemented by types in this
```
Pauan:
```suggestion
/// The trait is sealed: it can only be implemented by types in this
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+impl<'a> BlobContents for &'a str {
+    fn into_jsvalue(self) -> JsValue {
+        JsValue::from_str(self)
+    }
+}
+
+impl<'a> Sealed for &'a [u8] {}
+impl<'a> BlobContents for &'a [u8] {
+    fn into_jsvalue(self) -> JsValue {
+        let array = unsafe { js_sys::Uint8Array::view(self) };
```
Pauan: This can just be replaced with `js_sys::Uint8Array::from(self)`

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+            properties.type_(mime_type);
+        }
+
+        let parts = js_sys::Array::of1(&val);
+        let inner = web_sys::Blob::new_with_u8_array_sequence_and_options(&parts, &properties);
+
+        Blob::from(inner.unwrap_throw())
+    }
+
+    /// Like `new`, but allows specifying the media type (also known as *content type* or *mime
```
Pauan:
```suggestion
    /// Like `new`, but allows specifying the MIME type (also known as *content type*) of the `Blob`.
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+        let b: &web_sys::Blob = self.as_ref();
+        Blob::from(b.slice_with_f64_and_f64(start, end).unwrap_throw())
+    }
+
+    /// The number of bytes in the Blob/File.
+    pub fn size(&self) -> u64 {
+        safe_f64_to_u64(self.inner.size())
+    }
+
+    /// The typed *media type* (also known as *content type* or *mime type*) of the `File` or
```
Pauan:
```suggestion
    /// The statically typed MIME type (also known as *content type*) of the `File` or
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// `Blob`.
+    #[cfg(feature = "mime")]
+    pub fn mime_type(&self) -> Result<mime::Mime, mime::FromStrError> {
+        self.raw_mime_type().parse()
+    }
+
+    /// The *media type* (also known as *content type* or *mime type*) of the `File` or
```
Pauan:
```suggestion
    /// The raw MIME type (also known as *content type*) of the `File` or
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// `Blob` as a `String`.
```
Pauan:
```suggestion
    /// `Blob`.
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+    pub fn raw_mime_type(&self) -> String {
+        self.inner.type_()
+    }
+}
+
+impl<'a> From<&'a str> for Blob {
```
Pauan: I'm not so sure about these impls, I think they're covered by `Blob::new`.

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+/// `Blob`s can be created directly from `&str`, `&[u8]`, and `js_sys::ArrayBuffer`s using the
+/// `From::from` implementation (e.g. `let b: Blob = "test".into()`), or using the
+/// `new_with_options` constructor.
+#[derive(Debug, Clone)]
+pub struct Blob {
+    inner: web_sys::Blob,
+}
+
+impl Blob {
+    fn new(val: JsValue, mime_type: Option<&str>) -> Blob {
```
Pauan: This is missing a `pub fn new` for `Blob` (but there is one for `File`).

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+    {
+        Self::_new(
+            name,
+            contents.into_jsvalue(),
+            mime_type,
+            last_modified_since_epoch,
+        )
+    }
+
+    fn _new(
```
Pauan: This should be renamed to `new_` so it doesn't hide unused warnings.

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+    ///
+    /// `contents` can be from one of `&str`, `&[u8]`, or others.
+    pub fn new<T>(name: &str, contents: T) -> File
+    where
+        T: BlobContents,
+    {
+        Self::_new(name, contents.into_jsvalue(), None, None)
+    }
+
+    /// Like `File::new`, but allows customizing the last modified time and the media type (also
```
Pauan:
```suggestion
    /// Like `File::new`, but allows customizing the last modified time and the MIME type (also
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// known as *content type* or *mime type*)
```
Pauan:
```suggestion
    /// known as *content type*)
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+impl From<File> for Blob {
+    fn from(file: File) -> Self {
+        file.inner
+    }
+}
+
+// utility methods
+// ===============
+
+/// Numbers above 2^54 cannot always be represented exactly as an f64, javascript's native numeric
```
Pauan:
```suggestion
/// JavaScript only has `f64`, which has a maximum accurate integer size of`2^53 - 1`. So we use this to safely convert from larger integers to `f64`.
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,338 @@
...
+/// type. See
+/// [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER)
+fn safe_u64_to_f64(number: u64) -> f64 {
+    // Max integer stably representable by f64
+    // todo use js_sys::Number::MAX_SAFE_INTEGER once stable
+    const MAX_SAFE_INTEGER: u64 = 9007199254740991; // (2^53 - 1)
+    assert!(
+        number <= MAX_SAFE_INTEGER,
+        format!(
```
Pauan: I'd like to avoid the formatting code, but I'm not entirely sure how.

crates/file/src/file_reader.rs
derekdreery: Do you mean to avoid pulling in `std::fmt` so it can be pruned if no user code uses it?

Pauan: I mean that it shouldn't be using `format!`, since that pulls in all of the formatting infrastructure.

Implementing `std::fmt::Display` or `std::fmt::Debug` is fine, since Rust will remove it if it's not used.

It's the act of *using* `std::fmt::Display` or `std::fmt::Debug` that pulls it in. Normally that's done by using `format!` or `println!`.

derekdreery: In that case we also want to exclude the `write!` macro used in `impl Display for FileReadError`.

crates/file/src/blob.rs
derekdreery: I spent a bit of time trying to come up with the correct name for mime/media/content type. I think I went for media based on [its wikipedia page](https://en.wikipedia.org/wiki/Media_type) saying that it's now called media. However MIME might be better if it's more recognisable. Which would you prefer?

crates/file/src/blob.rs
derekdreery: Yes they're just there because they can be (i.e. there is a sensible conversion). I can remove them if you like?

crates/file/src/blob.rs
Pauan: MIME type is used in the W3C spec and MDN, I think it's *far* more familiar and commonly used than media type.

And since the property is called `mime`, I think it should put MIME first, but I don't have a particularly strong opinion on it.

crates/file/src/blob.rs
Pauan: I'd rather us not have them. We can always add them later, but we can't remove them after we've added them.

crates/file/src/blob.rs
derekdreery: What you suggest sounds good :).

crates/file/src/file_reader.rs
derekdreery: You are correct. :)

derekdreery:
Ok I think I've addressed all the issues from the review, so it should be ready for review again.

crates/file/src/blob.rs
```diff
@@ -0,0 +1,309 @@
...
+pub struct File {
+    // the trick here is that we know the contents of `inner` are a file, even though that type
+    // information is not stored. It is the same trick as is used in `web_sys`.
+    inner: Blob,
+}
+
+impl File {
+    /// Create a new `File` with the given name and contents.
+    ///
+    /// `contents` can be from one of `&str`, `&[u8]`, or others.
```
Pauan:
```suggestion
    /// `contents` can be `&str`, `&[u8]`, or `ArrayBuffer`.
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,309 @@
...
+    pub fn new<T>(name: &str, contents: T) -> File
+    where
+        T: BlobContents,
+    {
+        Self::new_(name, contents.into_jsvalue(), None, None)
+    }
+
+    /// Like `File::new`, but allows customizing the last modified time and the MIME type (also
```
Pauan: Docs order should match the argument order, mentioning MIME type first and then last modified time.

crates/file/src/blob.rs
```diff
@@ -0,0 +1,309 @@
...
+        File::from(inner)
+    }
+
+    /// Gets the file name.
+    pub fn name(&self) -> String {
+        let f: &web_sys::File = self.as_ref();
+        f.name()
+    }
+
+    /// Gets the duration since the epoch that the file was last modified.
```
Pauan:
```suggestion
    /// Gets the duration since the Unix epoch that the file was last modified.
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,309 @@
...
+    ///
+    /// `std::time::Duration` is a low level type, use a crate like
+    /// [`chrono`](https://docs.rs/chrono/0.4.10/chrono/) to work with more user-friendly
```
Pauan:
```suggestion
    /// [`chrono`](https://docs.rs/chrono/0.4.10/chrono/) to work with a more user-friendly
```

crates/file/src/blob.rs
```diff
@@ -0,0 +1,309 @@
...
+    /// representations of time.
+    pub fn last_modified_since_epoch(&self) -> Duration {
+        let f: &web_sys::File = self.as_ref();
+        Duration::from_millis(safe_f64_to_u64(f.last_modified()))
+    }
+
+    /// Create a new `File` from a part of this `File`.
```
Pauan:
```suggestion
    /// Create a new `File` from a sub-part of this `File`.
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,234 @@
...
+        fn drop(&mut self) {
+            if !ReadyState::from(self.reader.ready_state()).is_done() {
+                self.reader.abort();
+            }
+        }
+    }
+
+    /// Asynchronously converts `blob` into a text string and then passes it to the `callback`.
+    ///
+    /// If the returned `FileReader` is dropped before the callback is invoked, the read will be
```
Pauan:
```suggestion
    /// If the returned `FileReader` is dropped before the callback is called, the read will be
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,234 @@
...
+            callback,
+            |value| value.as_string().unwrap_throw(),
+            |reader, blob| reader.read_as_text(blob).unwrap_throw(),
+        )
+    }
+
+    /// Asynchronously converts the `blob` into a base64 encoded `data:` URL and then passes it to the `callback`.
+    /// encoded data url, and then run the callback.
+    ///
+    /// If the returned `FileReader` is dropped before the callback is invoked, the read will be
```
Pauan:
```suggestion
    /// If the returned `FileReader` is dropped before the callback is called, the read will be
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,234 @@
...
+        read(
+            blob,
+            callback,
+            |value| value.as_string().unwrap_throw(),
+            |reader, blob| reader.read_as_text(blob).unwrap_throw(),
+        )
+    }
+
+    /// Asynchronously converts the `blob` into a base64 encoded `data:` URL and then passes it to the `callback`.
+    /// encoded data url, and then run the callback.
```
Pauan:
```suggestion
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,234 @@
...
+            blob,
+            callback,
+            |value| value.as_string().unwrap_throw(),
+            |reader, blob| reader.read_as_data_url(blob).unwrap_throw(),
+        )
+    }
+
+    /// Asynchronously converts the `blob` into an array buffer and then passes it to the `callback`.
+    ///
+    /// If the returned `FileReader` is dropped before the callback is invoked, the read will be
```
Pauan:
```suggestion
    /// If the returned `FileReader` is dropped before the callback is called, the read will be
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,234 @@
...
+            blob,
+            callback,
+            |value| value.dyn_into::<js_sys::ArrayBuffer>().unwrap_throw(),
+            |reader, blob| reader.read_as_array_buffer(blob).unwrap_throw(),
+        )
+    }
+
+    /// Asynchronously converts the `blob` into a `Vec<u8>` and then passes it to the `callback`.
+    ///
+    /// If the returned `FileReader` is dropped before the callback is invoked, the read will be
```
Pauan:
```suggestion
    /// If the returned `FileReader` is dropped before the callback is called, the read will be
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,234 @@
...
+    /// returned by the read and the type the callback expects, and `read_fn` is the method that
+    /// runs the async read on the `FileReader`.
+    fn read<T, CF, EF, RF>(blob: &Blob, callback: CF, extract_fn: EF, read_fn: RF) -> FileReader
+    where
+        CF: FnOnce(Result<T, FileReadError>) + 'static,
+        EF: Fn(JsValue) -> T + 'static,
+        RF: Fn(&web_sys::FileReader, &web_sys::Blob) + 'static,
+    {
+        // we need to be able to run the FnOnce, while proving to the compiler that it can only run
+        // once. The easiest way is to `take` it out of an Option. The `Arc` and `Mutex` are
```
Pauan:
```suggestion
        // once. The easiest way is to `take` it out of an Option. The `Rc` and `RefCell` are
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,234 @@
...
+
+        let reader = web_sys::FileReader::new().unwrap_throw();
+
+        let load_reader = reader.clone();
+        let error_reader = reader.clone();
+        // Either the load listener or the error listener will be called, never both (so FnOnce is
+        // ok).
+        let load_listener = EventListener::new(&reader, "load", move |_event| {
+            let result = extract_fn(load_reader.result().unwrap_throw());
+            load_callback.borrow_mut().take().unwrap_throw()(Ok(result));
```
Pauan:
```suggestion
            let callback = load_callback.borrow_mut().take().unwrap_throw();
            callback(Ok(result));
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,234 @@
...
+            let exception = error_reader.error().unwrap_throw();
+            let error = match exception.name().as_str() {
+                "NotFoundError" => FileReadError::NotFound(exception.message()),
+                "NotReadableError" => FileReadError::NotReadable(exception.message()),
+                "SecurityError" => FileReadError::Security(exception.message()),
+                // This branch should never be hit, so returning a less helpful error message is
+                // less of an issue than pulling in `format!` code.
+                _ => throw_str("unrecognised error type"),
+            };
+            error_callback.borrow_mut().take().unwrap_throw()(Err(error));
```
Pauan:
```suggestion
            let callback = error_callback.borrow_mut().take().unwrap_throw();
            callback(Err(error));
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,234 @@
...
+    pub async fn read_as_text(blob: &Blob) -> Result<String, FileReadError> {
+        let (sender, receiver) = futures_channel::oneshot::channel();
+        let _reader = super::callbacks::read_as_text(blob, |result| {
+            sender.send(result).unwrap_throw();
+        });
+        receiver.await.unwrap_throw()
+    }
+
+    /// Returns the contents of `blob` as a base64 encoded `data:` URL.
+    /// encoded data url.
```
Pauan:
```suggestion
```

derekdreery:
Ready for review again.

I've experimented with changing the last_modified type to `SystemTime`, since a) it seems more semantically correct, and b) it allows times before the epoch. Just for your consideration, happy to change it back if you prefer.

Pauan:
@derekdreery I'm torn. On the one hand, it does seem semantically correct. On the other hand, the last modified time must always be *after* the Unix Epoch, never before. I guess some runtime checking is acceptable.

crates/file/src/blob.rs
```diff
@@ -147,46 +149,67 @@ pub struct File {
...
+    /// );
+    /// assert_eq!(example_file.name(), "motivation.txt".into());
+    /// assert_eq!(example_file.raw_mime_type(), "text/plain".into());
+    /// ```
     pub fn new_with_options<T>(
         name: &str,
         contents: T,
         mime_type: Option<&str>,
-        last_modified_since_epoch: Option<Duration>,
+        last_modified: Option<SystemTime>,
```
Pauan:
```suggestion
        last_modified_time: Option<SystemTime>,
```

crates/file/src/blob.rs
```diff
@@ -147,46 +149,67 @@ pub struct File {
...
     where
         T: BlobContents,
     {
-        Self::new_(
-            name,
-            contents.into_jsvalue(),
-            mime_type,
-            last_modified_since_epoch,
-        )
+        Self::new_(name, contents.into_jsvalue(), mime_type, last_modified)
```
Pauan:
```suggestion
        Self::new_(name, contents.into_jsvalue(), mime_type, last_modified_time)
```

crates/file/src/blob.rs
```diff
@@ -147,46 +149,67 @@ pub struct File {
...
     }
 
+    /// Private constructor.
     fn new_(
         name: &str,
         contents: JsValue,
         mime_type: Option<&str>,
-        last_modified_since_epoch: Option<Duration>,
+        last_modified: Option<SystemTime>,
```
Pauan:
```suggestion
        last_modified_time: Option<SystemTime>,
```

crates/file/src/blob.rs
```diff
@@ -147,46 +149,67 @@ pub struct File {
...
     ) -> Self {
         let mut options = web_sys::FilePropertyBag::new();
         if let Some(mime_type) = mime_type {
             options.type_(&mime_type);
         }
 
-        if let Some(last_modified) = last_modified_since_epoch {
-            options.last_modified(safe_u128_to_f64(last_modified.as_millis()));
+        if let Some(last_modified) = last_modified {
```
Pauan:
```suggestion
        if let Some(last_modified) = last_modified_time {
```

Pauan:
Nevermind, I just checked and times before the epoch are valid, so `SystemTime` seems definitely correct.

crates/file/src/blob.rs
```diff
@@ -202,17 +225,29 @@ impl File {
...
+    ///
+    /// ```rust
+    /// use chrono::prelude::*;
+    /// use gloo_file::File;
+    ///
+    /// // Just create a dummy `gloo::file::File` for demonstration purposes.
+    /// let example_file = File::new("test_file.txt", "<almost empty contents>");
+    /// let date: DateTime<Utc> = example_file.last_modified().into();
+    /// ```
+    pub fn last_modified(&self) -> SystemTime {
```
Pauan:
```suggestion
    pub fn last_modified_time(&self) -> SystemTime {
```

crates/file/src/blob.rs
```diff
@@ -227,7 +262,7 @@ impl File {
             &self.name(),
             blob.into(),
             mime_type.as_ref().map(|s| s.as_str()),
-            Some(self.last_modified_since_epoch()),
+            Some(self.last_modified()),
```
Pauan:
```suggestion
            Some(self.last_modified_time()),
```

crates/file/tests/web.rs
```diff
@@ -45,6 +45,22 @@ fn png_file() {
...
+#[wasm_bindgen_test]
+fn modified_since() {
+    use chrono::Utc;
+    use std::time::{Duration, SystemTime};
+
+    let file = File::new("test.txt", "");
+    let now: SystemTime = Utc::now().into();
+    // Check the file was created less than 1ms ago. This should be plenty of time.
+    assert!(
+        file.last_modified()
```
Pauan:
```suggestion
        file.last_modified_time()
```

crates/file/src/file_reader.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+}
+
+#[cfg(feature = "futures")]
+pub mod futures {
+    use crate::{Blob, FileReadError};
+    use wasm_bindgen::UnwrapThrowExt;
+
+    /// Returns the contents of `blob` as a text string.
+    pub async fn read_as_text(blob: &Blob) -> Result<String, FileReadError> {
```
Pauan: By the way, I just realized that these `async` functions capture the lifetime for their argument, which isn't good. So they should be implemented like this instead:

```rust
pub fn read_as_text(blob: &Blob) -> impl Future<Output = Result<String, FileReadError>> {
    let (sender, receiver) = futures_channel::oneshot::channel();

    let reader = super::callbacks::read_as_text(blob, move |result| {
        sender.send(result).unwrap_throw();
    });

    async move {
        let output = receiver.await.unwrap_throw();
        drop(reader);
        output
    }
}
```

crates/file/src/blob.rs
derekdreery: `last_modified` would match the underlying webidl, but I agree with you that adding `_time` is more readable.

derekdreery:
Would you like me to squash changes down to a single commit?

crates/file/src/blob.rs
Pauan: With gloo we don't try to match 1-to-1, that's what web-sys is for.

Pauan:
@derekdreery I don't care much either way, so it's up to you.

derekdreery:
I squashed :)

Pauan:
Thanks a lot! I'm really happy with how this turned out.

derekdreery:
woop woop :)

derekdreery:
I wonder if @alexcrichton would mind commenting on the use of `SystemTime` for the `last_modified_time` return type, since if I recall correctly he wrote the `std::time` RFC.

alexcrichton:
I haven't really reviewed this much, but I would agree that using libstd types where possible is the best path to take.

derekdreery:
@alexcrichton really I was just asking if it seems reasonable to use `SystemTime` in wasm32-unknown-unknown, or not. Sounds like you think it's fine. :)

PR #113, opened 2020-02-29
First draft of RFC for websockets API.

derekdreery:
This is a first draft of a websocket API. I don't think it's ready to merge yet, but I wanted to start the conversation.

[rendered](https://github.com/derekdreery/gloo/edit/ws-rfc/rfcs/000-websocket-api.md?pr=%2Frustwasm%2Fgloo%2Fpull%2F113)

### DIFF
--- a/rfcs/000-websocket-api.md
+++ b/rfcs/000-websocket-api.md
@@ -29,7 +29,100 @@ modules: `callback` and `futures`. The `callback` API will be a thin wrapper aro
 providing some type safety and reducing boilerplate where possible. The `futures` API will use
 rust async concepts (`Stream + Sync`) to make it as easy and risk-free to use as possible.
 
-### `callback` api
+### The shared API
+
+```rust
+
+/// An incoming websocket message.
+pub enum Message {
+    /// Message was in the binary variation.
+    Binary(Vec<u8>),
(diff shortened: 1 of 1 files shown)


derekdreery:
Can someone explain to me how to link the rendered document without it being tied to a specific blob?

derekdreery:
ping

Lemmih:
I would like to assist. The draft of the API looks complete enough for an experimental implementation. What would be the best next steps for this?

hamza1311:
I just saw this. I would suggest using an API based on channels/streams as that's consistent with other websocket crates and, imo, is more "rust-y". See [this implementation of such an API in `reqwasm`](https://github.com/hamza1311/reqwasm/blob/master/src/websocket.rs) as an example.

Lemmih:
The `futures` API from the proposal is consistent with the API from `reqwasm`, right?

hamza1311:
Seems that way. I somehow missed the figured part. This API looks good.

hamza1311 (changes requested):
I left a few comments. Generally, it looks good.

hamza1311:
Another thing: @derekdreery, can you create an issue for this RFC with the changes. See #136 for details

PR #125, opened 2021-05-04
Add wrappers for web storage API

### LINKED ISSUE #124: Add wrapper for `localStorage` and `sessionStorage`
## Summary

Add wrappers for browser's storage interface. These will wrap around `localStorage` and `sessionStorage` and provide an idiomatic Rust API.

## Motivation

It's hard to deal with `web-sys` directly and using storage in browser is a common case.

## Detailed Explanation

We provide a trait with default implementations for dealing with `web_sys::Storage`. The `web_sys::Storage` instance to be used is provided by the `raw` method which must be implemented by the implementers of the trait. We use `serde` to serialize and deserialize data to and from JSON which is stored in the browser-provided storage.

```rust
pub struct JsError {
    pub name: String,
    pub message: String,
    /// Holds the value returned by calling `toString` on the error. This is used by `Display` impl.
    js_to_string: String,
}

#[derive(Debug, thiserror::Error)]
pub enum StorageError {
    #[error("{0}")]
    SerdeError(#[from] serde_json::Error),
    #[error("key {0} not found")]
    KeyNotFound(String),
    #[error("{0}")]
    JsError(JsError),
}

pub type Result<T> = std::result::Result<T, StorageError>;

pub trait Storage {
    /// Get the raw [`web_sys::Storage`] instance
    fn raw() -> web_sys::Storage;

    /// Get the value for the specified key
    fn get<T>(key: impl AsRef<str>) -> Result<T>
    where
        T: for<'de> Deserialize<'de>,
    { ... }

    /// Get all the stored keys and their values
    fn get_all<T>() -> Result<T>
    where
        T: for<'a> Deserialize<'a>,
    { ... }

    /// Insert a value for the specified key
    fn set<T>(key: impl AsRef<str>, value: T) -> Result<()>
    where
        T: Serialize,
    { ... }

    /// Remove a key and it's stored value
    fn delete(key: impl AsRef<str>) { ... }

    /// Remove all the stored data
    fn clear() { ... }

    /// Get the number of items stored
    fn length() -> u32 { ... }
}

pub struct LocalStorage;

impl Storage for LocalStorage {
    fn raw() -> web_sys::Storage { ... }
}

pub struct SessionStorage;

impl Storage for SessionStorage {
    fn raw() -> web_sys::Storage { ... }
}
```

### Alternative Approach 

Provide an API with a single struct that's holds on to a `web_sys::Storage` instance and performs operations with that.

```rust
pub enum Area {
    Local,
    Session,
}
pub struct Storage {
    storage: web_sys::Storage,
}

impl Storage {
    pub fn new(area: Area) -> Self { ... }
}
```

## Unresolved Questions

None

hamza1311:
Fixes #124

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ gloo-timers = { version = "0.2.0", path = "crates/timers" }
 gloo-console-timer = { version = "0.1.0", path = "crates/console-timer" }
 gloo-events = { version = "0.1.0", path = "crates/events" }
 gloo-file = { version = "0.1.0", path = "crates/file" }
+gloo-storage = { version = "0.1.0", path = "crates/storage" }
 
 [features]
 default = []
--- a/crates/file/README.md
+++ b/crates/file/README.md
@@ -20,10 +20,14 @@
   <sub>Built with 🦀🕸 by <a href="https://rustwasm.github.io/">The Rust and WebAssembly Working Group</a></sub>
 </div>
 
+Provides wrappers for working with `Blob`s and `File`s from JavaScript.
 
-Working with files and blobs on the Web.
+A `File` is just a `Blob` with some extra data: a name and a last modified time.
 
-These APIs come in two flavors:
+In the File API, `Blob`s are opaque objects that lazily fetch their contained data when
+asked. This allows a `Blob` to represent some resource that isn't completely available, for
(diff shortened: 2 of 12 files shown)


jstarry (approved):
Looks great, ready for merge?

hamza1311:
> Looks great, ready for merge?

Sure, let's do it.

PR #126, opened 2021-05-06
Add wrappers for requestAnimationFrame

hamza1311:


crates/render/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+
+/// Calls browser's `requestAnimationFrame`. It is cancelled when the handler is dropped.
+///
+/// [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame)
+pub fn request_animation_frame<F>(callback: F) -> AnimationFrame
+where
+    F: Fn(f64) + 'static,
+{
+    let callback = move |v: JsValue| {
+        let time: f64 = v.as_f64().unwrap_or(0.0);
```
jstarry, line 33: The animation frame callback is only called once right? How about when it's called, we drop the closure and clear the render_id so that the `AnimationFrame` drop implementation becomes a no-op?

jstarry: > Also, trying to drop the callback and render_id early would just add more complexity. I don't understand how we could do that.

Typically it's done using something like `Rc<RwLock<Option<...>>>`. You could then clone the `Rc` into the closure to get a mutable reference to the inner option

hamza1311: Right, but I'm not sure if it's worth the overhead considering there has to be code in `Drop` impl. I mentioned this above:

> We would still need to keep the code in drop because what if `AnimationFrame` is dropped before the callback is called.

jstarry: Most times the callback will not be dropped before being called, by skipping the `cancel_animation_frame` we save some relatively slow browser api calls and free up any memory that is captured in the user's callback

hamza1311: Done that.

hamza1311:
@jstarry, merged your PR. Ready for merge now?

jstarry (approved):

Issue #119, opened 2020-06-05
Need a maintainer


Pauan:
Since I am no longer contributing to Rust, somebody else will need to take over as the maintainer of this project.

If anybody is interested in that, please contact @alexcrichton (who is a Rust Wasm WG Core member).

tomseattle:
Pauan, Would love the chance to help as maintainer. Sent email to Alex.

Stupremee:
@jstarry I would also be ready to help you out with maintaining this project.

tomseattle:
@jstarry Sounds good, lets coordinate on some of the open issues and also some of the documentation. It would good if we can co-maintain if thats possible,  thanks

MATRIXKOO:
Still, need a maintainer?

jstarry:
Great to have all this interest, I think the real need here is contributors. It would be great to assess the state of current open PR's and issues first. If merging PR's and publishing crates becomes the bottleneck, we can share maintainer responsibilities.

MATRIXKOO:
Contributor sounds good too.

jstarry:
@alexcrichton would you mind updating my role from what I think is "contributor" to "maintainer" so that I can another maintainer to this repo?

alexcrichton:
Sure, I've bumped you to "admin" since you were already "maintainer"

jstarry:
Much appreciated, thanks!

hamza1311:
@alexcrichton, can you give me the permission to modify secrets in this repo. I need the permissions for #140 (see the [failing CI](https://github.com/rustwasm/gloo/runs/2926113072?check_suite_focus=true#step:4:10))

alexcrichton:
If it's ok I'm trying to have a"hands-off" approach to this crate so I don't have to personally be on the hook for things. I think @jstarry should be an admin, so they should have all the requisite privileges for changing things?

hamza1311:
> If it's ok I'm trying to have a"hands-off" approach to this crate so I don't have to personally be on the hook for things. I think @jstarry should be an admin, so they should have all the requisite privileges for changing things?

If I interpreted their message on discord correctly, them and I have same permissions that's why I pinged you. If @jstarry can do it, I'd be fine with that. Again, I apologize for pinging you. If you're fine with it, I wouldn't mind having admin permissions for this repo

alexcrichton:
Hm if an admin can't add another admin perhaps this repository should be trasnferred out of the rustwasm org?

[...]

PR #129, opened 2021-06-11
[RFC] Add FileReaderSync API

DesmondWillowbrook:
[Rendered](https://github.com/DesmondWillowbrook/gloo/blob/master/rfcs/003-add-filereadersync.md)


DesmondWillowbrook:
We can also implement `std::io::Read` for `Blob`/`File` using `FileReaderSync`.
However, the implementation involves using `readAsArrayBuffer` and such, which makes it look like a mid-level API.

My implementation:

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=914a904f05f2cf7b2dcf87deda7a7eea

DesmondWillowbrook:
cc @alexcrichton @fitzgen  @Pauan

Pauan:
Note that [I no longer contribute to Rust](https://github.com/rustwasm/gloo/issues/119), and I don't know who the current maintainers of Gloo are.

Also note that `FileReaderSync` can only be used in Workers... and it can't run in parallel, so it's not very useful.

Instead you should just use [the `Future` version of `FileReader`](https://github.com/rustwasm/gloo/blob/f4859029380f2b33634efe46ac127c6e652892f5/crates/file/src/file_reader.rs#L144-L221), which already exists:

```rust
use gloo_file::FileReadError;
use gloo_file::futures::read_as_array_buffer;

async fn foo(blob: Blob) -> Result<(), FileReadError> {
    let array_buffer = read_as_array_buffer(&blob).await?;
    // ...
    Ok(())
}
```

DesmondWillowbrook:
Understood.

And yes, I am aware of the fact. 
My current use was porting over a program to WebAssembly (which was not designed with async in mind). I couldn't find an easy way to `block_on()` my file reading function since `wasm_bindgen_futures` only featured a `block_on()` which returned no arguments.

The easiest way, then, to get a Reader to read a `File` passed from JS was to put the instantiation code in a Web Worker and use `FileReaderSync`.

That is the reason I thought of polishing this API and putting it in `gloo-files`, as a low-level which did not require dealing with JS types.

Pauan:
> I couldn't find an easy way to block_on() my file reading function since wasm_bindgen_futures only featured a block_on() which returned no arguments.

There is no `block_on` in the browser, because the browser does not provide *any* way to do synchronous blocking for Promises. It is a limitation in the browser, so there's nothing we can do about it. Browsers are fundamentally designed to be async, they *really* do not like sync code.

> The easiest way, then, to get a Reader to read a File passed from JS was to put the instantiation code in a Web Worker and use FileReaderSync.

You can read a file synchronously in a Worker, but Workers are really limited in their API: for example, Workers cannot access the DOM. So in order to do anything useful you have to send it back to the main thread, and sending messages is always asynchronous. So unless your application is *very* simple, you won't be able to run it entirely in a Worker.

DesmondWillowbrook:
Yes, you're very much right.

Async in the browser is very much preferred, and the fact that WebAssembly does not feature any sort of threading makes blocking on async in WebAssembly modules difficult.

Yes, that is true. My case was a file processing application which only really required the input of a file, some parameters, and a method to output JSON. It was a _port_ of a command-line application, thus it went just fine. 😀

Pauan:
> My case was a file processing application which only really required the input of a file, some parameters, and a method to output JSON. It was a port of a command-line application, thus it went just fine.

Ah, okay, that does indeed seem like an example of a simple program which can actually run entirely in a Worker.

alexcrichton:
Sorry but I don't personally maintain gloo

hamza1311:
> Note that [I no longer contribute to Rust](https://github.com/rustwasm/gloo/issues/119), and I don't know who the current maintainers of Gloo are.

I think the CONTRIBUTION.md should be updated accordingly. I'm also not sure of the current members so in case anyone does, a PR would be great.

hamza1311:
Hey @DesmondWillowbrook, can you create an issue for this RFC. See #136 for details.

hamza1311:
Closing this in favor of #138

PR #133, opened 2021-06-23
Fix CI

hamza1311:


crates/file/src/blob.rs
```diff
@@ -335,6 +335,7 @@ fn safe_f64_to_u64(number: f64) -> u64 {
     if number > MAX_SAFE_INTEGER {
         throw_str("a rust number was too large and could not be represented in JavaScript");
     }
+    #[allow(clippy::float_cmp)] // TODO find a good solution?
```
hamza1311, line 338: Does anyone know of a better solution for this?

CC: @jstarry

DesmondWillowbrook: We could use [`js_sys::Number::is_safe_integer`](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Number.html#method.is_safe_integer), though that requires the overhead of converting to a `Number` and back.

Looks like [`js_sys::Number::MAX_SAFE_INTEGER`](https://rustwasm.github.io/wasm-bindgen/api/js_sys/struct.Number.html#associatedconstant.MAX_SAFE_INTEGER) is stable. And since we know the number is supposed to have no fractional part, we can use `as u64`? I believe `as u64's default behavior is to truncate the fractional part.

Does that make sense?

jstarry: Looks like [`fract`](https://doc.rust-lang.org/std/primitive.f64.html#method.fract) would work well. Source: https://stackoverflow.com/a/48500414

hamza1311: I switched to `fract` and used `js_sys::Number::MAX_SAFE_INTEGER` where needed.

PR #134, opened 2021-06-24
Add `doc` attibute to `cfg(feature=future)`

DesmondWillowbrook:
So it shows up in `cargo doc --open`.

Previously:
image
[docs link](https://docs.rs/gloo-file/0.1.0/gloo_file/index.html)

Now:
image

We would want to also add [`doc(cfg)`](https://doc.rust-lang.org/rustdoc/unstable-features.html#doccfg-recording-what-platforms-or-features-are-required-for-code-to-be-present) once it's stabilized, as it gives dedicated highlighting to features which should be enabled for a certain item.


DesmondWillowbrook:
Looks like the easiest fix for this is to run `cargo doc --all-features` to generate documentation. (since it enables futures and documents it as well.)

PR #147, opened 2021-07-26
FileReaderSync RFC Implementation

### LINKED ISSUE #138: [RFC] Add FileReaderSync API to gloo-file
## Summary

Add [`FileReaderSync`](https://developer.mozilla.org/en-US/docs/Web/API/FileReaderSync) API to `gloo-file`

## Motivation

To enable synchronous File reading from Web Workers.

## Detailed Explanation

Unlike the callback-based FileReader, this API is not callback-driven, which means that we can make it simpler, as:
- A RAII style struct for automatically cancelling the read is not required, which means we don't need a struct.

```rust
mod FileReaderSync {
    pub fn readAsArrayBuffer (blob: &Blob) -> Result<ArrayBuffer, FileReadError>
    // ... similar method signatures to FileReader, without the callbacks
}
```

Can consider transforming `FileReaderSync` into a _trait_ which is implemented on `Blobs` and such, which makes the API prettier, but less clear?
```rust
use gloo_file::FileReaderSync;

let file = get_file().readAsArrayBuffer();
let data = Uint8Array::new(file);
```

Prior art:
- [FileReaderSync MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/FileReaderSync)
- [FileReader implementation in Gloo](https://github.com/rustwasm/gloo/blob/master/crates/file/src/file_reader.rs)

## Unresolved Questions

Might look into adding a few more error-types to `FileReadError`, which are specific to particular types of reads, such as `EncodingError` present in [`readAsDataURL`](https://developer.mozilla.org/en-US/docs/Web/API/FileReaderSync/readAsDataURL).

Don't know whether they are in the scope of this PR.

DesmondWillowbrook:
Closes #138


crates/file/src/sync_file_reader.rs
```diff
@@ -0,0 +1,29 @@
...
+/// cancelled.
+pub fn read_as_text (blob: &Blob) -> Result<String, FileReadError>
+{
+	let fr = FileReaderSync::new().unwrap_throw();
+	fr.read_as_text(blob.as_ref()).map_err(
+		|e| {
+			throw_str(&format!("{:?}", e));
+
+			let e= e.as_string().unwrap_throw();
+			match e.as_ref () {
```
DesmondWillowbrook, line 20: @hamza1311
I'm having some trouble trying to log a JsValue. [`throw_str`](https://docs.rs/wasm-bindgen/0.2.74/wasm_bindgen/fn.throw_str.html) or [`throw_val`](https://docs.rs/wasm-bindgen/0.2.74/wasm_bindgen/fn.throw_val.html) from `wasm_bindgen` don't seem to be working, as browser tests output (this was interactive, as I haven't been able to get headless tests running, I'm on WSL):
```
failures:

---- web::sync_file_from_str output ----
    JS exception that was thrown:
        Error: `unwrap_throw` failed
        init/imports.wbg.__wbindgen_throw@http://127.0.0.1:8000/wasm-bindgen-test:799:15
        wasm_bindgen::throw_str::h70fa659b3cee3ef4@http://127.0.0.1:8000/wasm-bindgen-test_bg.wasm:wasm-function[2172]:0x8db5a
... really long stack trace
```
which looks to be the `unwrap_throw` on the `JsValue` passed to `map_err` by `read_as_text`.

I really just want to process the `JsValue` into a `FileReadError`, but looking at the `FileReader` implementation didn't help as it used callbacks and the exception API looked different.

DesmondWillowbrook:
Oh and currently I've just bodged the naming, re-exporting and public API of the crate to just get some tests running, so I will clean that up later on. Considering that `gloo_file` re-exports the `FileReader` methods verbatim, and that `FileReader` and `FileReaderSync` both have the same method names, we will have to make some calls on shuffling the methods around, or renaming all `FileReaderSync` methods to start with `sync`, eg. `sync_read_as_text`.

crates/file/src/sync_file_reader.rs
DesmondWillowbrook: cc @hamza1311

hamza1311: I'm sorry it took me so long to get back to you, been busy with things irl. I'm not really sure what's happening here. I've never experienced it. You can also use `gloo_console` to log. That would make the output look better too.

DesmondWillowbrook: That doesn't work. Though I think I know why the `throw_str` stuff isn't working either. Rather obvious in hindsight, `SyncFileReader` is only available in Web Workers and the context that the tests are being run in is clearly _not_ that.

PR #152, opened 2021-07-31
Remove the unnecessary copy in `Blob::new`

### LINKED ISSUE #150: `Blob::new(&[u8])` makes an unnecessary copy
The implementation of `BlobContents` for `&[u8]` uses `Uint8Array::from` to make a `Uint8Array` to pass to `Blob::new`. That method creates a copy of the slice, which is then immediately passed to `Blob::new`, which creates another copy of the `Uint8Array`. The first copy isn't necessary, because it can instead pass `Blob::new` a `Uint8Array` view into wasm memory. `Uint8Array::view` is unsafe, because the returned `Uint8Array` allows mutating the immutable buffer and has no lifetime, but it's fine to pass to `Blob::new`.

As a downside, this would require making `BlobContents::into_jsvalue` unsafe, because it would allow other code outside of `gloo` to invoke `Uint8Array::view` safely. I don't think it's really meant to be used outside of `gloo` anyway, though.

Liamolucko:
Use `Uint8Array::view` instead of `Uint8Array::from` to avoid making two copies of slices passed to `Blob::new`.

Fixes #150

### DIFF
--- a/crates/file/src/blob.rs
+++ b/crates/file/src/blob.rs
@@ -11,14 +11,14 @@ mod sealed {
 use sealed::Sealed;
 
 /// This trait is used to overload the `Blob::new_with_options` function, allowing a variety of
-/// types to be used to create a `Blob`. Ignore this, and use &[u8], &str, etc to create a `Blob`.
+/// types to be used to create a `Blob`. Ignore this, and use &\[u8], &str, etc to create a `Blob`.
 ///
 /// The trait is sealed: it can only be implemented by types in this
 /// crate, as this crate relies on invariants regarding the `JsValue` returned from `into_jsvalue`.
 pub trait BlobContents: Sealed {
     /// # Safety
     ///
-    /// For `&[u8]`, the returned `Uint8Array` must be modified,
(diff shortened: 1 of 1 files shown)


crates/file/src/blob.rs
```diff
@@ -189,7 +197,14 @@ impl File {
     where
         T: BlobContents,
     {
-        Self::new_(name, contents.into_jsvalue(), mime_type, last_modified_time)
+        Self::new_(
+            name,
+            // SAFETY: `Self::new_` doesn't modify the `Uint8Array`,
+            // or hold on to the reference past the end of the function call.
+            unsafe { contents.into_jsvalue() },
```
hamza1311, line 204: I think we should pass `contents` to `Self::new_`. Currently, someone could change `Self::new_` without realizing it this safety requirements. Also, that would reduce the number of places `unsafe` block is required.

hamza1311 (changes requested):

hamza1311 (approved):
LGTM

PR #154, opened 2021-09-03
Fix dir, dirxml macros in gloo-console

WorldSEnder:
Noticed the discussion on discord and thought I'd contribute a fix.

Gets rid of a repetition that is not allowed according to [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/console/dir), which only mentions one single argument. Adds two examples that get picked up by doctesting to verify the fix.


hamza1311 (approved):
LGTM. WIll merge once CI is green

PR #155, opened 2021-09-16
Utility crate for common `web_sys`/`js_sys` features

### LINKED ISSUE #72: Utility crate for common web_sys features
## Summary

I propose adding a crate that wraps common web_sys features with a cleaner API.

Examples:
- Wrap things like `Window`, `Document` and `History` in a way that handles failures automatically with `expect_throw`
- Provide shims to get or set the `value` of input etc elements
- Wrap `console::log` and `console::error` with a more flexible API

## Motivation

The `web_sys` api can be verbose; high-level crates needn't each provide their own wrappers.

## Detailed Explanation

Some common features of `web_sys` are verbose when used directly. It's likely that many `Gloo` crates will make use of them - let's settle on a general wrapping API higher-level crates can use. [Example](https://github.com/David-OConnor/seed/blob/master/src/util.rs)

Examples:
```rust
/// Convenience function to avoid repeating expect logic.
pub fn window() -> web_sys::Window {
    web_sys::window().expect_throw("Can't find the global Window")
}

/// Convenience function to access the web_sys DOM document.
pub fn document() -> web_sys::Document {
    window().document().expect_throw("Can't find document")
}

/// Convenience function to access the web_sys history.
pub fn history() -> web_sys::History {
    window().history().expect_throw("Can't find history")
}

/// Simplify getting the value of input elements; required due to the need to cast
/// from general nodes/elements to HTML__Elements.
pub fn get_value(target: &web_sys::EventTarget) -> String {
    if let Some(input) = target.dyn_ref::<web_sys::HtmlInputElement>() {
        return input.value();
    }
    if let Some(input) = target.dyn_ref::<web_sys::HtmlTextAreaElement>() {
        return input.value();
    }
    if let Some(input) = target.dyn_ref::<web_sys::HtmlSelectElement>() {
        return input.value();
    }
    "".into()
}

/// Similar to get_value
pub fn set_value(target: &web_sys::EventTarget, value: &str) {
    if let Some(input) = target.dyn_ref::<web_sys::HtmlInputElement>() {
        input.set_value(value);
    }
    if let Some(input) = target.dyn_ref::<web_sys::HtmlTextAreaElement>() {
        input.set_value(value);
    }
    if let Some(input) = target.dyn_ref::<web_sys::HtmlSelectElement>() {
        input.set_value(value);
    }
}
```

The original `web_sys` APIs aren't that clumsy - this might clutter up `Gloo` without providing much benefit. Perhaps we want to use pattern matching to handle failures of `web_sys::Window` etc instead of using `expect_throw`.

It's easy to create a collection of helper functions that are intended to be general, but spread over a wide range of uses, some of which may not be common. It's possible that they may cover a smaller set of use-cases than intended. Thin wrappers in general can be seen as obfuscating the original, well-known API. Why build a new API that causes confusion with the one it wraps, when it's not much better?

## Unresolved Questions

It's unclear what the scope of this should be. There are surely grey areas about what's appropriate here, and how cohesive/comprehensive it should be. What causes `web_sys::Window` etc to fail?

I'm certainly biased by my own use-cases. Do y'all think these are, in-fact useful, general wrappers that would see use in many/most `Gloo` crates?

I've only scratched the surface. What else should be added?

### LINKED ISSUE #153: Add a wrapper for the JavaScript `Error` type
Right now, it's very hard to extract information from `js_sys::Error`. `gloo` should provide a wrapper type, `JsError` (better name welcome) which provides this information in an easy-to-use fashion.

This should go in `gloo-utils` crate which should be introduced as part of #72

hamza1311:
This PR adds:

- Convenience functions for
  - `window`
  - `document`
  - `history`
- Ergonomic wrapper for `js_sys::Error` type

Fixes #72 
Fixes #153


crates/utils/src/errors.rs
```diff
@@ -0,0 +1,54 @@
...
+impl TryFrom<JsValue> for JsError {
+    type Error = NotJsError;
+
+    fn try_from(value: JsValue) -> Result<Self, Self::Error> {
+        match value.dyn_into::<js_sys::Error>() {
+            Ok(error) => Ok(JsError::from(error)),
+            Err(_) => Err(NotJsError),
+        }
+    }
+}
```
mc1098, line 39:
```suggestion
impl TryFrom<JsValue> for JsError {
    type Error = JsValue;

    fn try_from(value: JsValue) -> Result<Self, Self::Error> {
        value.dyn_into::<js_sys::Error>().map(JsError::from)
    }
}
```
The `NotJsError` doesn't really give you much and you lose the `JsValue` when calling `JsError::try_from`, chances are most of the time this won't matter but I think this change gives more flexibility?

hamza1311: How about keeping `NotJsValue` but giving it access to the `JsValue` and implement std `Error` for it? It can also be used with anyhow that way

mc1098: Sounds good to me :)

PR #171, opened 2021-11-23
Add BrowserHistory and BrowserLocation

futursolo:
The PR adds the initial `BrowserHistory` and `BrowserLocation` that was initially implemented for `yew-router`.

Comparing to `yew-router`'s implementation, it strips the following features:
- typed routes.
- base name handling.

### History leaking
(a.k.a: sometimes it's possible to cause the page to reload, or navigate away from the SPA with certain methods like `history.go(0)`)

After some consideration, I think it ultimately does not matter too much and is kind of hard to prevent.

- `history.go(0)` is equivalent to `location.reload()`.
   Which means the user should know what they are doing when they are calling this method with 0.
- `history.back()` or `history.go(-1)` (or `history.forward()` and `history.go(1)`)
   If the last state is created by `history.pushState`, then a `popstate` event will be fired.
   If the last state is on another page, then user will be navigated back to the last page.
   If this is the first state in the history queue, then nothing happens.
   In addition, there's no way to "peek" the next history which means we cannot really know what would happen unless it really happens.

JavaScript version of `history` does not patch / trying to prevent this either:

https://github.com/remix-run/history/blob/main/packages/history/index.ts#L521

### DIFF
--- a/crates/history/Cargo.toml
+++ b/crates/history/Cargo.toml
@@ -1,0 +1,44 @@
+[package]
+name = "gloo-history"
+version = "0.1.0"
+description = "Universal Session History"
+authors = ["Rust and WebAssembly Working Group"]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+repository = "https://github.com/rustwasm/gloo/tree/master/crates/history"
+homepage = "https://github.com/rustwasm/gloo"
+categories = ["api-bindings", "storage", "wasm"]
+
+[dependencies]
+wasm-bindgen = "0.2"
+gloo-utils = { version = "0.1.0", path = "../utils" }
+gloo-events = { version = "0.1.0", path = "../events" }
+serde = "1"
+serde_urlencoded = "0.7"
+serde-wasm-bindgen = "0.3.1"
+thiserror = "1.0"
+
+[dependencies.web-sys]
+version = "0.3"
+features = [
+    # "Attr",
+    # "Document",
+    "History",
+    # "HtmlBaseElement",
+    # "Event",
+    # "NamedNodeMap",
+    # "Url",
+    # "UrlSearchParams",
+    "Window",
+]
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3"
+serde = { version = "1", features = ["derive"] }
+
+# [dev-dependencies.web-sys]
+# version = "0.3"
+# features = [
+#     "HtmlHeadElement",
+# ]
--- a/crates/history/Cargo.toml
+++ b/crates/history/Cargo.toml
@@ -22,23 +22,12 @@ thiserror = "1.0"
 [dependencies.web-sys]
 version = "0.3"
 features = [
-    # "Attr",
-    # "Document",
     "History",
-    # "HtmlBaseElement",
-    # "Event",
-    # "NamedNodeMap",
-    # "Url",
-    # "UrlSearchParams",
     "Window",
+    "Location",
 ]
 
 [dev-dependencies]
 wasm-bindgen-test = "0.3"
 serde = { version = "1", features = ["derive"] }
-
-# [dev-dependencies.web-sys]
-# version = "0.3"
-# features = [
-#     "HtmlHeadElement",
-# ]
+gloo-timers = { version = "0.2.0", features = ["futures"], path = "../timers" }
--- a/crates/history/src/any.rs
+++ b/crates/history/src/any.rs
@@ -126,9 +126,9 @@ impl History for AnyHistory {
 impl Location for AnyLocation {
     type History = AnyHistory;
 
-    fn pathname(&self) -> String {
+    fn path(&self) -> String {
         let Self::Browser(self_) = self;
-        self_.pathname()
+        self_.path()
     }
 
     fn search(&self) -> String {
--- a/crates/history/src/any.rs
+++ b/crates/history/src/any.rs
@@ -113,14 +113,6 @@ impl History for AnyHistory {
         let Self::Browser(self_) = self;
         AnyLocation::Browser(self_.location())
     }
-
-    fn state<T>(&self) -> HistoryResult<T>
-    where
-        T: DeserializeOwned + 'static,
-    {
-        let Self::Browser(self_) = self;
-        self_.state()
-    }
 }
 
 impl Location for AnyLocation {
@@ -147,6 +139,14 @@ impl Location for AnyLocation {
     fn hash(&self) -> String {
         let Self::Browser(self_) = self;
         self_.hash()
+    }
+
+    fn state<T>(&self) -> HistoryResult<T>
+    where
+        T: DeserializeOwned + 'static,
+    {
+        let Self::Browser(self_) = self;
+        self_.state()
     }
 }
 
--- a/crates/history/src/any.rs
+++ b/crates/history/src/any.rs
@@ -1,0 +1,163 @@
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+
+use crate::browser::{BrowserHistory, BrowserLocation};
+use crate::error::HistoryResult;
+use crate::history::History;
+use crate::listener::HistoryListener;
+use crate::location::Location;
+
+/// A [`History`] that is always available under a [`Router`](crate::Router).
+#[derive(Clone, PartialEq)]
+pub enum AnyHistory {
+    Browser(BrowserHistory),
+}
+
+/// The [`Location`] for [`AnyHistory`]
+#[derive(Clone, PartialEq)]
+pub enum AnyLocation {
+    Browser(BrowserLocation),
+}
+
+impl History for AnyHistory {
+    type Location = AnyLocation;
+
+    fn len(&self) -> usize {
+        let Self::Browser(self_) = self;
+        self_.len()
+    }
+
+    fn go(&self, delta: isize) {
+        let Self::Browser(self_) = self;
+        self_.go(delta)
+    }
+
+    fn push(&self, route: impl Into<String>) {
+        let Self::Browser(self_) = self;
+        self_.push(route)
+    }
+
+    fn replace(&self, route: impl Into<String>) {
+        let Self::Browser(self_) = self;
+        self_.replace(route)
+    }
+
+    fn push_with_state<T>(&self, route: impl Into<String>, state: T) -> HistoryResult<()>
+    where
+        T: Serialize + 'static,
+    {
+        let Self::Browser(self_) = self;
+        self_.push_with_state(route, state)
+    }
+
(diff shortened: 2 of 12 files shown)


crates/history/src/any.rs
```diff
@@ -0,0 +1,163 @@
...
+pub enum AnyLocation {
+    Browser(BrowserLocation),
+}
+
+impl History for AnyHistory {
+    type Location = AnyLocation;
+
+    fn len(&self) -> usize {
+        let Self::Browser(self_) = self;
+        self_.len()
```
hamza1311, line 27: Let's make the job of the next person updating this code to include another `History` a easier (and make the code, imo, more readable) by using a `match` here.

crates/history/src/history.rs
```diff
@@ -0,0 +1,102 @@
...
+        self.go(1);
+    }
+
+    /// Loads a specific page in [`History`] with a `delta` relative to current page.
+    ///
+    /// See: <https://developer.mozilla.org/en-US/docs/Web/API/History/go>
+    fn go(&self, delta: isize);
+
+    /// Pushes a route entry with [`None`] being the state.
+    fn push(&self, route: impl Into<String>);
```
hamza1311, line 35:
```suggestion
    fn push(&self, route: impl AsRef<str>);
```

Avoid the owned `String` allocation

crates/history/src/history.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    /// Pushes a route entry with state.
+    ///
+    /// The implementation of state serialization differs between [`History`] types.
+    ///
+    /// For [`BrowserHistory`] and [`HashHistory`], state is serialised with [`serde_wasm_bindgen`] where as
+    /// [`MemoryHistory`] uses [`Any`](std::any::Any).
+    fn push_with_state<T>(&self, route: impl Into<String>, state: T) -> HistoryResult<()>
+    where
+        T: Serialize + 'static;
```
hamza1311, line 48: `serde` should be feature gated.

crates/history/src/listener.rs
```diff
@@ -0,0 +1,9 @@
+use std::rc::Rc;
+
+/// A History Listener to manage callbacks registered on a [`History`].
+///
+/// This Listener has the same behaviour as the [`EventListener`] from [`gloo`]
+/// that the underlying callback will be unregistered when the listener is dropped.
+pub struct HistoryListener {
```
hamza1311, line 7: This should have the [`must_use` attribute](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute).

crates/timers/src/future.rs
```diff
@@ -84,7 +84,7 @@ impl TimeoutFuture {
 ///
 /// # Example
 ///
-/// ```no_run
+/// ```compile_fail
```
hamza1311, line 87: Why modify timers here? Let's do it in a separate PR

hamza1311 (changes requested):
Looks fine for the most part. I've left a few comments (some of them apply everywhere but I've noted them at one place)

crates/history/src/history.rs
futursolo: This is actually deliberate.

This is due to routes(`Routable`) for `yew-router` is usually declared as an enum, and can implement `Into<String>` easily, but cannot implement `AsRef<str>` easily.

I guess maybe we can do `Into<Cow<'_, str>>` instead?

crates/history/src/history.rs
futursolo: Making serde to be optional actually will not only lose `location.state` but also `location.query`. Which I consider to be significant enough to have it on by default.

FWIW, `gloo-console` and `gloo-storage` also have serde to be a non-optional dependency.

crates/timers/src/future.rs
futursolo: This is purely to make CI happy or else CI wouldn't pass.

https://github.com/rustwasm/gloo/runs/4295589107?check_suite_focus=true

crates/history/src/history.rs
hamza1311: `gloo-storage` needs it and it should also be optional for `gloo-console`. We can feature gate but enable that feature by default.

crates/history/src/history.rs
hamza1311: > I guess maybe we can do `Into<Cow<'_, str>>` instead?

That would be good. That way we can avoid the allocation when possible

crates/history/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+name = "gloo-history"
+version = "0.1.0"
+description = "Universal Session History"
+authors = ["Rust and WebAssembly Working Group"]
+edition = "2018"
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+repository = "https://github.com/rustwasm/gloo/tree/master/crates/history"
+homepage = "https://github.com/rustwasm/gloo"
+categories = ["api-bindings", "storage", "wasm"]
```
hamza1311, line 11: This crate isn't related to storage

crates/history/src/browser.rs
```diff
@@ -0,0 +1,383 @@
...
+        self.inner
+            .go_with_delta(delta as i32)
+            .expect_throw("failed to call go.")
+    }
+
+    fn push<'a>(&self, route: impl Into<Cow<'a, str>>) {
+        let url = route.into();
+        self.inner
+            .push_state_with_url(&JsValue::NULL, "", Some(&url))
+            .expect("failed to push state.");
```
hamza1311, line 60:
```suggestion
            .expect_throw("failed to push state.");
```

Please do the same for other `expect`s too

crates/history/src/browser.rs
```diff
@@ -0,0 +1,383 @@
...
+                    let history_clone = history.clone();
+
+                    // Listens to popstate.
+                    LISTENER.with(move |m| {
+                        let mut listener = m.borrow_mut();
+
+                        *listener = Some(EventListener::new(&window, "popstate", move |_| {
+                            history_clone.notify_callbacks();
+                        }));
+                    });
```
hamza1311, line 234:
```suggestion
                {
                    let history = history.clone();

                    // Listens to popstate.
                    LISTENER.with(move |m| {
                        let mut listener = m.borrow_mut();

                        *listener = Some(EventListener::new(&window, "popstate", move |_| {
                            history.notify_callbacks();
                        }));
                    });
                }
```
Style changes

crates/history/src/listener.rs
```diff
@@ -0,0 +1,17 @@
+use std::fmt;
+use std::rc::Rc;
+
+/// A History Listener to manage callbacks registered on a [`History`].
+///
+/// This Listener has the same behaviour as the [`EventListener`] from [`gloo`]
+/// that the underlying callback will be unregistered when the listener is dropped.
+#[must_use]
```
hamza1311, line 8:
```suggestion
#[must_use = "the listener is removed when `HistoryListener` is dropped"]
```

hamza1311 (commented):
Looks good for the most part, just a few comments

crates/history/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+    "Location",
+]
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3"
+serde = { version = "1", features = ["derive"] }
+gloo-timers = { version = "0.2.0", features = ["futures"], path = "../timers" }
+
+[features]
+serialize = ["serde", "serde_urlencoded", "serde-wasm-bindgen", "thiserror"]
```
hamza1311, line 36: Let's split this up into two features `query` and `state`. We can avoid bringing in an unnecessary crate that way

hamza1311 (commented):
I forgot to add this in the previous review...

crates/history/src/any.rs
```diff
@@ -0,0 +1,203 @@
+use std::borrow::Cow;
+
+#[cfg(feature = "serde")]
+use serde::de::DeserializeOwned;
+#[cfg(feature = "serde")]
+use serde::Serialize;
```
hamza1311, line 6: There's no `serde` feature now.

crates/history/src/browser.rs
```diff
@@ -0,0 +1,385 @@
...
+
+use gloo_events::EventListener;
+use gloo_utils::window;
+#[cfg(feature = "serde")]
+use serde::de::DeserializeOwned;
+#[cfg(feature = "serde")]
+use serde::Serialize;
+use wasm_bindgen::{JsValue, UnwrapThrowExt};
+
+#[cfg(feature = "serde")]
```
hamza1311, line 14: Again, no `serde` features. This needs to be updated according to what's used

crates/history/src/browser.rs
```diff
@@ -0,0 +1,385 @@
...
+        HistoryListener { _listener: cb }
+    }
+
+    fn location(&self) -> Self::Location {
+        BrowserLocation::new(self.clone())
+    }
+}
+
+impl Default for BrowserHistory {
+    fn default() -> Self {
```
hamza1311, line 203: I saw your messages on the Yew discord advising against directly calling it. Should this be `pub(crate)` function instead of `Default` implementation? That way, it won't be part of the public API

crates/history/src/history.rs
```diff
@@ -0,0 +1,121 @@
+use std::borrow::Cow;
+
+#[cfg(feature = "serde")]
+use serde::Serialize;
+
+#[cfg(feature = "serde")]
```
hamza1311, line 6: More `serde` features...

hamza1311 (commented):
Only the references to `serde` feature which was renamed stuck out to me. Other than that, this looks good to me

crates/history/src/any.rs
futursolo: When an optional dependency is added, a feature with the same name is implicitly registered.

> Optional dependencies implicitly define a feature of the same name as the dependency. This means that the same cfg(feature = "gif") syntax can be used in the code, and the dependency can be enabled just like a feature such as --features gif (see Command-line feature options below).

https://doc.rust-lang.org/cargo/reference/features.html

crates/history/src/browser.rs
futursolo: I didn’t recommend it due to it is not a good practice inside components and users should use hooks instead. Not that default implementation itself has any issue.

I think an updated implementation will no longer expose History interface via hooks, so I am fine with default implementation.

In addition, it trips up clippy if you have a new method, but no default implementation.

crates/history/src/any.rs
hamza1311: TIL

hamza1311:
LGTM. 

I'm going to go ahead and merge this so it's not a blocking factor for more history-related work in both Gloo and Yew. Should anything needs to be patched up, it can be done in another PR

PR #174, opened 2021-11-26
Fixes `console_dbg!` and `console!` expression output. Bold src info.

Madoshakalaka:
```rust
    let a = 1;
    gloo_console::console_dbg!("123", a, a + 1);
    gloo_console::console_dbg!(String::from("123"));
    gloo_console::console!(b());

    let b: JsString = "123".into();
    gloo_console::console!(b);

    gloo_console::console!({let c: JsString = "cc".into(); c});
```

image

### DIFF
--- a/crates/console/src/console_dbg.rs
+++ b/crates/console/src/console_dbg.rs
@@ -6,16 +6,15 @@
 #[macro_export]
 macro_rules! console {
     () => {
-        $crate::log!(::std::format!("[{}:{}]", ::std::file!(), ::std::line!()));
+        $crate::log!(
+            ::std::format!("%c[{}:{}] ", ::std::file!(), ::std::line!()),
+            "font-weight: bold",
+        );
     };
     ($val:expr $(,)?) => {
-        // Use of `match` here is intentional because it affects the lifetimes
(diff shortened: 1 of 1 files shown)


crates/console/src/console_dbg.rs
```diff
@@ -6,16 +6,15 @@
...
+        );
     };
     ($val:expr $(,)?) => {
-        // Use of `match` here is intentional because it affects the lifetimes
-        // of temporaries - https://stackoverflow.com/a/48732525/1063961
-        match $val {
-            tmp => {
-                $crate::log!(::std::format!("[{}:{}] {} =", ::std::file!(), ::std::line!(), ::std::stringify!($val)), &tmp);
-                tmp
-            }
```
Madoshakalaka, line 18: This is no longer necessary, existed for two reasons:
1. expressed called twice 
2. align lifetimes of multiple expressions 

1 won't happen anymore because now both macros only run the expression once and bind them to a variable.
2 is not necessary. It's only needed in cases like assert_eq.

Madoshakalaka:
@hamza1311 had an extra comma that failed the tests. If you can approve the CI, it should pass now

hamza1311 (approved):

PR #177, opened 2021-12-04
Hash-based History type & Unified Location.

futursolo:
This pull request consists of the following changes:
- `HashHistory` for hash-based history access.
- `Location` is now a struct and its content will be stable after created by `History::location()`.
- `Location` now have an `id` that is unique to itself.
- States now uses downcasting in all occasions.


hamza1311:
It seems there's some code duplication between the two histories, life for notifications. Can we avoid that?

hamza1311 (approved):
LGTM

PR #180, opened 2022-01-02
Add `gloo-worker` crate

hamza1311:
Migrate `yew-agent` to `gloo-worker`

The code in this PR is moved from the [Yew repository](https://github.com/yewstack/yew) as of commit yewstack/yew@6f6519d. All the credits go to original code authors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ gloo-render = { version = "0.1.0", path = "crates/render" }
 gloo-console = { version = "0.2.1", path = "crates/console" }
 gloo-utils = { version = "0.1.1", path = "crates/utils" }
 gloo-history = { version = "0.1.0", path = "crates/history" }
+gloo-worker = { path = "crates/worker" }
 
 [features]
 default = []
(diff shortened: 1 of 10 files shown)


futursolo:
This looks fine to me for migrating from `yew-agent` to `gloo-worker`.

However, as we don't have non-worker based "agents" anymore, I think the implementation could be revised to be more specialised to implement workers:
- Do we need Reach on the worker itself?
    If we remove Reach from the worker trait, then a worker can be spawned as any type and the type can be specified at the time it is created. (this also means that we can merge `PublicWorker` and `PrivateWorker` together.)
- Currently, the path of a worker is not dynamic (specified in name_of_resource). If we make it possible to specify the path at the time a worker is created, will it be easier to handle relative paths? (as well as whether a worker is defined as a ES module.)

hamza1311:
I'm going to merge this then.  
I agree with the changes. Do you want to PR those changes? If not, I think we should release this as 0.1 right now. The implementation revisions could come later (I don't really have the time to make these changes myself right now)

futursolo:
> Do you want to PR those changes?

Not at anytime soon. I already have a lot on my plate at this moment.

PR #200, opened 2022-03-18
Worker v2

futursolo:
This pull request simplifies Workers to have 1 single type.

This consists of the following changes:

1. Unifies `PublicWorker` and `PrivateWorker`
2. Added a `WorkerSpawner` to supersede `Worker::bridge()` with a builder pattern to spawn workers. WorkerSpawner always spawn a new worker.
3. Removed `WorkerLink` and opted for `WorkerScope` directly.
4. `WorkerBirdge::fork()` which supports the former PublicWorker usage. It can "fork" the bridge with a different callback and HandlerId.

This pull request is feature complete.
However, I have modified some associated types on messages which could cause the main implementation of Worker to not be able to be tree shaked automatically. I will test this as I update the yew-agent crate and turn this into a pull request once I have the new agent implementation ready.


hamza1311:
It would be nice if you could also add tests for worker implementation. Assuming that's possible, of course.

futursolo:
It's actually relatively cumbersome to compile workers at the moment. However, the next version of trunk has introduced worker support which we may have a better chance after the release of the next version of Trunk.

droogmic:
Can this be completed now that trunk supports workers?

droogmic:
FYI: I had a problem moving to v2, because the resource path was missing this prefix:
https://github.com/droogmic/gloo/blob/master/crates/worker/src/worker/mod.rs#L85-L89

futursolo:
> FYI: I had a problem moving to v2, because the resource path was missing this prefix:
> https://github.com/droogmic/gloo/blob/master/crates/worker/src/worker/mod.rs#L85-L89

I think this has been fixed in 64a08e0 as it now uses the URL of current page as base.
Could you please try again?

If it does not work, could you provide the URL of the page and the path of the worker that is passed to .spawn()?

droogmic:
[5a0bd6](https://github.com/rustwasm/gloo/pull/200/commits/c5a0bd6905a6a1fee19736ffc9ac5e2037735ae8) looks incredible, thank you!

droogmic:
> > FYI: I had a problem moving to v2, because the resource path was missing this prefix:
> > https://github.com/droogmic/gloo/blob/master/crates/worker/src/worker/mod.rs#L85-L89
> 
> I think this has been fixed in [64a08e0](https://github.com/rustwasm/gloo/commit/64a08e0dcde4ccb1b256180f3b34529591584fd7) as it now uses the URL of current page as base. Could you please try again?
> 
> If it does not work, could you provide the URL of the page and the path of the worker that is passed to .spawn()?

yup, this works, checked in https://github.com/a-b-street/osm2lanes/pull/214/commits/f1eff7da19257ab6275b00db2f648080a07b4bfc

Now I need to figure out how to integrate rkyv as an alternative encoding to bincode (bincode doesn't play nice with serde's flatten)

droogmic:
> [5a0bd6](https://github.com/rustwasm/gloo/pull/200/commits/c5a0bd6905a6a1fee19736ffc9ac5e2037735ae8) looks incredible, thank you!

ah, I spoke too soon. This still depends on serde. Maybe there is another serde-compatible binary coding crate that doesn't have the same problems.

futursolo:
> serde-compatible binary coding

You should be able to use `serde_json` or `serde-wasm-bindgen`.
The message format accepts any type that supports structured clone.

droogmic:
Thanks @futursolo, I got this working end to end now here: https://github.com/a-b-street/osm2lanes/pull/214

I will keep this updated with your changes.

droogmic (approved):

crates/worker/src/codec.rs
```diff
@@ -0,0 +1,38 @@
...
+
+    /// Decode a message to a type
+    fn decode<O>(input: JsValue) -> O
+    where
+        O: for<'de> Deserialize<'de>;
+}
+
+/// Default message encoding with [bincode].
+#[derive(Debug)]
+pub struct Bincode {}
```
hamza1311, line 20: Should bincode be the default, with no way to opt out? Postcard recently hit 1.0, which is no_std so it should have a smaller footprint in the binary. I think we should allow users to customize it. A simple feature flag (enabled by default) will be enough here. Those who want to customize can implement `Codec` manually

What do you think?

crates/worker/src/native_worker.rs
```diff
@@ -0,0 +1,69 @@
...
+                CODEC: Codec,
+                F: 'static + Fn(T)
+            {
+                let handler = move |message: MessageEvent| {
+                    let msg = CODEC::decode(message.data());
+                    handler(msg);
+                };
+                let closure = Closure::wrap(Box::new(handler) as Box<dyn Fn(MessageEvent)>);
+                self.set_onmessage(Some(closure.as_ref().unchecked_ref()));
+                // Memory leak?
```
hamza1311, line 51: Yes. This is a memory leak, the closure is never freed.

Is it possible to store the closure somewhere? If this is a hot path, the memory leak(s) can add up. How many times will this be called?

crates/worker/src/registrar.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// Sets a new message encoding.
+    pub fn encoding<C>(&self) -> WorkerRegistrar<W, C>
+    where
+        C: Codec,
+    {
+        WorkerRegistrar {
+            _marker: PhantomData,
+        }
+    }
```
hamza1311, line 45: This seems confusing. Why not just call `new` again? It's not like there's any state that needs to be kept. They both seem to be doing the same thing.

examples/markdown/src/bin/example_markdown_app.rs
```diff
@@ -0,0 +1,41 @@
...
+        .spawn("/example_markdown_worker.js");
+
+    bridge.send(MARKDOWN_CONTENT.to_owned());
+
+    spawn_local(async move {
+        bridge.send(MARKDOWN_CONTENT.to_owned());
+
+        // We need to hold the bridge until the worker resolves.
+        let promise = Promise::new(&mut |_, _| {});
+        let _ = JsFuture::from(promise).await;
```
hamza1311, line 39: I don't really understand this part. Shouldn't it be instant? Or do we need the tick from the micro task queue, after the promise is resolved?

examples/markdown/src/bin/example_markdown_test_server.rs
```diff
@@ -0,0 +1,19 @@
+#![cfg(not(target_arch = "wasm32"))]
```
hamza1311, line 1: Is this server only needed for test?

It's confusing how to use the example. Can you please add a README file, explaining how to run the example and the purpose of this server and it's relation with `trunk serve`?

hamza1311 (commented):
Thanks for this PR and all the amazing work that you've done. Sorry it look me a little while to get to this.

Since this is a complete rewrite, I think we should have a migration guide (not exactly sure where it would go). What do you think? I think a page on the website or a markdown file in this directory would do the job.

futursolo (commented):
I don't know where the migration guide should reside in either.

But as it is completely rewritten, it might be better to simply advise users to re-implement their agent like what's advised in the `yew-router` 0.16 migration guide.

hamza1311 (approved):
Thanks again for the PR!

huntc:
Thanks also from me! I think this is a great improvement.

PR #207, opened 2022-04-07
Fixed broken links to some crates

chapuzzo:
Those links were broken, just adding plurals fixes them all 😊


hamza1311:
Thanks for the PR

PR #214, opened 2022-04-20
A proposal for an IndexedDB wrapper

derekdreery:
This PR adds an indexedDB wrapper to `gloo_storage`. The main design decisions are:

 - `IdbRequest`s are wrapped as futures. An alternative would be to provide both callback-based and futures-based versions as in `gloo_file`, but when using IDB you have so many requests that this quickly becomes cumbersome.
 - There are different error types for different situations. I'm half-way through unifying the errors as much as possible. An alternative would be a single error type for the whole lib, or a type-erased error (`anyhow::Error` or `Box<dyn Error>`)
 
IndexedDB is fiendish to use because of things like transaction auto-commit and the request API, but I think that this wrapper makes it a lot easier to use, while still giving full zero-overhead access to the underlying web APIs.

Opening for feedback as I will have to park this for a few weeks (day job). I really hope it, or some other idb wrapper, will land.


derekdreery:
The example uses @Pauan 's `dominator` library as the DOM layer just to make rendering easier. I would like to do some more work to separate out the view from the IDB code so it's easier to read for someone not familiar with signals.

hamza1311:
Is this API ready for review? If so, can you post what the public API looks like?

Side note: I've never had to use IndexedDB so someone more knowledgeable should also review it

derekdreery:
Hi, sorry for the slow reply.

It is ready to be reviewed for overall strategy, but some details still need ironing out, things like function/type names and error types.

I will upload a document explaining design decisions are also copy its contents into this PR.

derekdreery:
Also pinging @c410-f3r

derekdreery:
I've made some improvements and push latest.

IMO the best way to review is to look at
 - the markdown content in this PR (and the README for gloo_storage), and
 - generated API docs for gloo_storage
 - (only after the first 2) the source

derekdreery:
:bell:

hamza1311:
Hi, sorry for the late response. I've personally never used IndexedDB so I can't comment much on the API. Can you comment what the public API looks like so it's easier for someone to review it, without having to look through the implementation?

Just to clarify: I'm fine with adding support for IndexedDB in gloo-storage. I can review the implementation when the PR is ready for merge. I just want someone more knowledgeable to review the public API.

derekdreery:
Cool, when I get chance I will put the API in the top comment.

PR #215, opened 2022-04-24
Add query() method

### LINKED ISSUE #198: gloo net add `.query` convenience method for `Request`
## Summary

add `.query` convenience method for `Request`

## Motivation

makes sending query string parameters more ergonomic

## Detailed Explanation

right now, we must manually embed the query string into the request url, which isn't as ergonomic for an http library as, for example, [`reqwest`](https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.query) with which you can do `.query(&[("key", "value")])`

i don't think there are drawbacks, refer to [`reqwest source`](https://docs.rs/reqwest/latest/src/reqwest/async_impl/request.rs.html#347) for implementation

## Unresolved Questions

burrscurr:
This commit adds a query() method to Request for conveniently adding query parameters (see https://github.com/rustwasm/gloo/issues/198). Internally, the query string is tracked using a wrapper around `web_sys::UrlSearchParams`.

The current implementation always appends to the query parameters added so far. It also preserves query parameters specified in the initial URL.

Open Questions / Thoughts:

 - Should `QueryParams` be pub? Perhaps the API would benefit from improvements before exposing it.
 - Should there be a method to replace the query parameters (similar to `headers()` replacing the headers and `header()` adding a header?
 - To preserve initial query parameters, the current implementation creates two `web_sys::Request`s, which feels pretty hacky. Haven't found a better way though.
 - In general, implementation would be easier if the URL passed to create a `Request` would be parsed directly there. However, this could fail and would require to change the return value to a `Result`.

I also would like to add some tests, but I did not quite get the test setup running, at least for now.

Closes #198.


crates/net/src/http/query.rs
```diff
@@ -0,0 +1,34 @@
...
+    raw: web_sys::UrlSearchParams,
+}
+
+impl Default for QueryParams {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl QueryParams {
```
hamza1311, line 15: Please add the following:
- `fn from_raw(raw: web_sys::UrlSearchParams) -> Self`
- `delete` method
- `get`method
- An iterator implementation. You can move `UncheckedIter`  from headers module to gloo-utils and also use it here.

crates/net/src/http/query.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub fn append(&self, name: &str, value: &str) {
+        self.raw.append(name, value)
+    }
+}
+
+impl fmt::Display for QueryParams {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.raw.to_string())
+    }
+}
```
hamza1311, line 34: What is the output of this? This seems like what should be a Debug implementation, not Display implementation

crates/net/src/http/mod.rs
```diff
@@ -119,6 +123,43 @@ impl Request {
...
+    /// It is possible to append the same parameters with the same name multiple times, so
+    /// `.query([("a", "1"), ("a", "2")])` results in the query string `a=1&a=2`.
+    ///
+    /// # Examples
+    ///
+    /// The query parameters can be passed as many different types:
+    ///
+    /// ```
+    /// # use std::collections::HashMap;
+    /// # use gloo_net::http::Request;
```
hamza1311, line 138:
```suggestion
    /// use std::collections::HashMap;
    /// use gloo_net::http::Request;
    ///
```

hamza1311 (changes requested):

hamza1311:
Can you also add test case(s)?

crates/net/src/http/query.rs
burrscurr: Its the query string without `?`, so for instance `a=1&b=2`.

crates/net/src/http/query.rs
burrscurr: I can add the first three methods. `UrlSearchParams` has no direct interface for iterating over the parameters, which seems to be similar to `web_sys::Headers`. Why/how does the iterator implementation for `Headers` ([here](https://docs.rs/gloo-net/latest/src/gloo_net/http/headers.rs.html#70)) work?

hamza1311: There is a JS iterator implementation that `Headers` uses

crates/net/src/http/query.rs
hamza1311: If you can add a more readable debug implementation, that would be great. Something like the output of `Vec<(&str, &str)>` should be good

crates/net/src/http/query.rs
hamza1311: #217 can help with the 3rd point. Headers exposes `js_sys::Iterator` which we can consume from Rust

burrscurr: I have added the missing methods and the iterator implementation. I temporarily moved `UncheckedIter` to the `http` module, after #217 is merged I will rebase.

hamza1311: #217 has been merged. You can rebase now

crates/net/tests/http.rs
```diff
@@ -115,3 +115,66 @@ async fn fetch_binary() {
...
+
+#[wasm_bindgen_test]
+fn query_params_escape() {
+    let params = QueryParams::new();
+    params.append("a", "1");
+    assert_eq!(params.to_string(), "a=1".to_string());
+
+    params.append("key", "ab&c");
+    assert_eq!(params.to_string(), "a=1&key=ab%26c");
+}
```
hamza1311, line 180: Can you move these tests to `mod tests` under `query.rs`? It would be similar to the tests added for `UncheckedIter` in #217

crates/net/src/http/mod.rs
```diff
@@ -119,6 +123,44 @@ impl Request {
...
+    ///     .query(slice_params)
+    ///     .query(vec_params)
+    ///     .query(map_params);
+    /// // Result URL: /search?key=value&a=3&b=4&key=another_value
+    /// # }
+    /// ```
+    pub fn query<'a, T, V>(self, params: T) -> Self
+    where
+        T: IntoIterator<Item = (&'a str, V)>,
+        V: ToString,
```
hamza1311, line 156: Is it to possible to make this `AsRef<str>`? We can avoid the clone that way/

crates/net/src/http/mod.rs
```diff
@@ -119,6 +123,44 @@ impl Request {
         self
     }
 
+    /// Append query parameters to the url, given as `(name, value)` tuples. Values can be of any
+    /// type that implements [`ToString`].
+    ///
+    /// It is possible to append the same parameters with the same name multiple times, so
+    /// `.query([("a", "1"), ("a", "2")])` results in the query string `a=1&a=2`.
+    ///
```
hamza1311, line 131:
```suggestion
    ///
    /// This method preserves the query parameters set as part of URL in [`Request::new`]
    ///
```

hamza1311 (commented):
Looks good. Just these couple of minor comments and rebase to fix the conflicts left

crates/net/src/http/mod.rs
burrscurr: I chose `ToString` because it allows to pass other types than `&str`/`String` and I don't see a way to preserve this with `AsRef<str>`. But I agree that the additional clone is not optimal.

hamza1311: `AsRef<str>` should also allow that. I believe any type that can be converted to `&str` will fall under that

burrscurr: When replacing `ToString` with `AsRef<str>` I get the following error:

```
error[E0277]: the trait bound `{integer}: AsRef<str>` is not satisfied
  --> src/http/mod.rs:150:12
   |
15 |     .query(vec_params)
   |      ----- ^^^^^^^^^^ the trait `AsRef<str>` is not implemented for `{integer}`
   |      |
   |      required by a bound introduced by this call
   ```

Which fails at the docstring example

Comment (burrscurr) [reply crates/net/src/http/mod.rs:158]: Or am I misunderstanding your suggestion?

Comment (hamza1311) [reply crates/net/src/http/mod.rs:158]: The error message makes the allocation explicit. Converting a number to string requires creating a String. We should keep that explicit

Comment (burrscurr) [reply crates/net/src/http/mod.rs]: Its `AsRef<str>` now, I've adjusted the doc example.

File: crates/net/src/http/query.rs [hunk 7]
```diff
@@ -0,0 +1,104 @@
...
+            .collect()
+    }
+
+    /// Remove all occurrences of a parameter from the query string.
+    pub fn delete(&self, name: &str) {
+        self.raw.delete(name)
+    }
+
+    /// Iterate over (name, value) pairs of the query parameters.
+    pub fn entries(&self) -> impl Iterator<Item = (String, String)> {
```
hamza1311, line 58: It should be `iter` but that's not blocking the PR. I'll rename it later if you don't want to

hamza1311 (approved):
Thanks for the PR. I'll merge it once the CI is green

crates/net/src/http/query.rs
burrscurr: Changed. Thanks for reviewing 👍🏻

burrscurr:
@hamza1311 Any idea why this test from `gloo_history` is failing? I cannot reproduce it locally.

hamza1311:
I don't know why. @futursolo added those so maybe they can have a look

futursolo:
Sorry, I am not sure why this is failing.

However, the failing test is not consistent and `window.history` is asynchronous so it's possible that the navigation didn't complete at the time of assertion if GitHub Actions is slow.

You can try to increase timeout for the other failing test and see if that helps.

hamza1311:
Thanks for the PR!

burrscurr:
Thanks for the review and the relentless work for `gloo`, `yew` etc.

PR #221, opened 2022-05-28
Worker: support rkyv in additon to bincode

droogmic:
I ran into issue https://github.com/bincode-org/bincode/issues/245 in bincode, which will seemingly not be fixed https://github.com/serde-rs/serde/issues/1346 without the implementation of a new stricter flatten attribute in serde.

Although I could remove `#[serde(flatten)]` from my type, but that only sidesteps the problem.

It would be best if gloo_worker supports something other than serde+bincode, so this attempts to add rkyv.

### DIFF
--- a/crates/worker/Cargo.toml
+++ b/crates/worker/Cargo.toml
@@ -18,7 +18,6 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
 anymap2 = "0.13"
-bincode = "1"
 gloo-console = { path = "../console", version = "0.2" }
 gloo-utils = { path = "../utils", version = "0.1" }
 js-sys = "0.3"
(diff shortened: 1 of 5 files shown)


droogmic:
# Help needed

I don't think I know rust well enough to know how to do this without advice.

What I tried to do is add rkyv as a feature, which then enables different trait bounds through `cfg`. Part way through the implementation, particularly when adding the `SerDe` helper trait, I began to wonder if macros weren't a better way to switch between trait bounds.

In it's current state, it doesn't compile. I don't know how to add the [`rkyv::Deserialize`](https://docs.rs/rkyv/latest/rkyv/trait.Deserialize.html) trait bound in with the correct generic type parameters.

https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features
The documentation hints that mutually exclusive features may best be done using different crates or runtime switching, so I am wondering if there is a better design where the current crate API needs a breaking change to do this properly?

droogmic:
There is also https://github.com/rustwasm/gloo/pull/200, which will significantly conflict with this effort.

PR #223, opened 2022-06-11
Remove the private worker refcell

huntc:
It should be possible to drop a private bridge from within its handler as it is in a public bridge.

When using private workers, if a bridge is dropped from within its handler then an interior mutability issue occurs. This is because a refcell was being used and the handler ref is borrowed while called. Then dropping within the handler also attempts a mutable borrow which then causes a panic.

The change here is to replace the refcell. Instead, the bridge callback is owned by the bridge's handler. We also take care to set the handler to an empty closure as the first part of dropping the bridge, thereby guaranteeing no further calls on the callback having dropped it.

I also note https://github.com/rustwasm/gloo/pull/200, which I think also avoids using the refcell. However, this PR is one that could be merged fairly quickly such that we can get a maintenance release out perhaps.

### DIFF
--- a/crates/worker/src/worker/private.rs
+++ b/crates/worker/src/worker/private.rs
@@ -4,10 +4,8 @@ use crate::{
 };
 use queue::Queue;
 use serde::{Deserialize, Serialize};
-use std::cell::RefCell;
 use std::fmt;
 use std::marker::PhantomData;
-use std::rc::Rc;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 thread_local! {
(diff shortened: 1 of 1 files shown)


futursolo (approved):

huntc:
I'm now wondering if my change is causing a memory leak of listeners which I'm observing in my app. Prior to my change, interior mutability was used to explicitly drop the callback when the bridge is dropped. However, I now see that `set_onmessage_closure` of `WorkerExt` makes a call `closure.forget()`... thus it appears to be the bridge's responsibility to drop it. I'm no longer dropping it then. Is someone able to confirm my understanding?

huntc:
https://github.com/rustwasm/gloo/pull/200 may have fixed my problem - will test.

hamza1311:
This was released in workers 0.1.2

huntc:
> This was released in workers 0.1.2

Thanks for letting us know. Users of 0.1.2 should revert to 0.1.1 with the original issue, or upgrade to 0.2.0. The 0.2.0 workers are much nicer anyhow, and fairly easy to migrate. 

My apologies though for inconvenience this may have caused. I had no idea that the closure was being forgotten in other parts of the codebase.

hamza1311:
I had released this change today.  I didn't look at this thread before releasing though.

There may be a case where this change is wanted and it's not like this is a security risk so I'll leave the release up

huntc:
> I had released this change today. I didn't look at this thread before releasing though.
> 
> There may be a case where this change is wanted and it's not like this is a security risk so I'll leave the release up

Ok. But for those not using the full coordinates eg 0.1, they’ll potentially get this release, and it will lead to a memory leak for private workers. The problem can be identified by the browser’s profiler and observing a continual increase in listeners.

PR #225, opened 2022-06-23
Implement std Error trait for JsError

### LINKED ISSUE #216: Implement std::error::Error for gloo_utils::errors::JsError
## Summary

Implement trait `std::error::Error` for `gloo_utils::errors::JsError`.

## Motivation

Many crate and utility function (e.g., `Context` in `anyhow` crate) requires the bound `std::error::Error`. However, gloo's `JsError` does not implement `Error`, though it is meant to representing an error.

## Detailed Explanation

The implementation should be simple. A blanket implementation should be good enough, as `std::error::Error` provides default implementation for trait functions.

```rust
impl std::error::Error for JsError {}
```

Implement this trait will be a *minor* breaking change, as describe in [here](https://github.com/rust-lang/rfcs/blob/721f2d74cc4daf76f3e49d58fbc6ded55d545e45/text/1105-api-evolution.md). However, the possibility of breaking existing code should be very small.

Once we implement this trait, removing it will be a *major* breaking change.

chmnchiang:
Implement `std::error::Error` for `gloo_utils::errors::JsError`.
Closes #216.

### DIFF
--- a/crates/utils/src/errors.rs
+++ b/crates/utils/src/errors.rs
@@ -76,3 +76,5 @@ impl fmt::Debug for JsError {
             .finish()
     }
 }
+
+impl std::error::Error for JsError {}


hamza1311:
Thanks for the PR!

PR #230, opened 2022-06-27
Add missing feature flags to gloo-net

### LINKED ISSUE #229: gloo-net: `websocket` feature is missing Serde dependency if used with `default-features = false`
### Describe the Bug

https://github.com/rustwasm/gloo/pull/222 introduced a dependency on `Serde` and `wasm-bindgen/serde-serialize` when using the `websocket` feature flag. The Cargo.toml file was not updated to reflect this.

arctic-alpaca:
Should fix #229

### DIFF
--- a/crates/net/Cargo.toml
+++ b/crates/net/Cargo.toml
@@ -55,6 +55,8 @@ websocket = [
     "pin-project",
     "futures-core",
     "futures-sink",
+    "wasm-bindgen/serde-serialize",
+    "serde",
 ]
 # Enables the HTTP API
 http = [


hamza1311:
Tests failures are unrelated. I don't see how not adding those features effect anything (since it compiles with websocket feature) but it doesn't hurt to add them. Thanks for the patch!

arctic-alpaca:
The issue was compiling gloo-net with `default-features=false` and the websocket feature. By default, all features are enabled. The  `json` feature pulls in the required dependencies for the websocket feature and hides the issue.

hamza1311:
Released in gloo-net 0.2.3 :tada:

PR #231, opened 2022-06-28
[rfc] Implement an ObjectUrl wrapper

WorldSEnder:
## Rationale

The [`createObjectUrl`](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL) API is offered by the browser to create URLs from Blobs that are too large to represent as data urls. The resulting url is along the lines of `blob:[origin]://[uuid]`. There's a counter part [`revokeObjectURL`](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL) that releases the URL again. Rust lends itself to an abstraction that automatically revokes the url when it is not need any longer.

## The API

A `struct ObjectUrl` abstracts the allocation of a specific ObjectUrl. It uses reference counting to revoke the Url after the last user drops its use. It can be created from a `Blob`, `File` or directly from a `web_sys::Blob`. A `Deref` implemention allows access to the stringly-typed url under-the-hood.

### DIFF
--- a/crates/file/src/object_url.rs
+++ b/crates/file/src/object_url.rs
@@ -1,0 +1,83 @@
+use crate::{Blob, Sealed};
+use std::{ops::Deref, rc::Rc};
+
+use wasm_bindgen::UnwrapThrowExt;
+use web_sys::Url;
+
+struct ObjectUrlAllocation {
+    url: String,
+}
+
+/// This trait is used to overload the `Url::create_object_url` function, allowing a variety of
+/// types to be used to create a [`ObjectUrl`]. Ignore this, and use [`Blob`] or [`File`] to create one.
+///
+/// The trait is sealed: it can only be implemented by types in this
+/// crate, as this crate relies on invariants regarding the `JsValue` returned from `into_jsvalue`.
+///
+/// [`File`]: crate::File
+pub trait ObjectUrlTarget: Sealed {
+    fn create_new_object_url(&self) -> String;
+}
+
+impl ObjectUrlTarget for Blob {
+    fn create_new_object_url(&self) -> String {
+        Url::create_object_url_with_blob(self.as_ref()).unwrap_throw()
+    }
+}
+
+impl Sealed for web_sys::Blob {}
+impl ObjectUrlTarget for web_sys::Blob {
+    fn create_new_object_url(&self) -> String {
+        Url::create_object_url_with_blob(self).unwrap_throw()
+    }
+}
+
+// Note: some browsers support Url::create_object_url_with_source but this is deprecated!
+// https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL#using_object_urls_for_media_streams
+
+impl ObjectUrlAllocation {
+    fn allocate_new_object_url(target: &dyn ObjectUrlTarget) -> Self {
+        let url = target.create_new_object_url();
+        Self { url }
+    }
+}
+
+impl Drop for ObjectUrlAllocation {
+    fn drop(&mut self) {
+        web_sys::Url::revoke_object_url(&self.url).unwrap_throw();
+    }
+}
+
+/// A resource wrapper around [`URL.createObjectURL`] / [`URL.revokeObjectURL`].
+///
+/// A [`Blob`], in particular a [`File`], can be converted to a short URL representing its data with the above methods.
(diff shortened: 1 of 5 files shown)


crates/file/src/object_url.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl ObjectUrl {
+    /// Create a new ObjectUrl from a [`Blob`] or [`File`].
+    ///
+    /// [`File`]: crate::File
+    pub fn new(target: &dyn ObjectUrlTarget) -> Self {
+        Self {
+            inner: Rc::new(ObjectUrlAllocation::allocate_new_object_url(target)),
+        }
+    }
```
hamza1311, line 74: Why not just have `From` impls? `From<Blob>`, `From<File>`, etc should be fine. I think it's better to use Rust's traits instead of creating our own, unless we have to.

A macro may be used to make implementing traits easier. Rust's orphan rules will ensure that no more traits can be implemented, effectively sealing it

crates/file/src/object_url.rs
```diff
@@ -0,0 +1,83 @@
...
+/// An [`ObjectUrl`] can be cheaply cloned and shared and revokes the underlying URL when the last reference is dropped.
+///
+/// Note that multiple urls can be created for the same blob, without being guaranteed to be de-deduplicated.
+///
+/// [`URL.createObjectURL`]: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL
+/// [`URL.revokeObjectURL`]: https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL
+/// [`File`]: crate::File
+#[derive(Clone)]
+pub struct ObjectUrl {
+    inner: Rc<ObjectUrlAllocation>,
```
hamza1311, line 63: `Rc<str>` instead of (effectively) `Rc<String>`? I think we can avoid the double heap allocation

WorldSEnder: `web_sys::Url::create_object_url_with_blob` already returns a heap allocated `String`. And I'm actually not sure how to convert a `JsString` to a `Rc<str>`, anyway. Got any pointers for that, then perhaps use a custom `#[wasm_bindgen] fn create_object_url_with_blob(&Blob) -> Result<JsString, JsValue>`?

crates/file/src/object_url.rs
WorldSEnder: BlobContents

crates/file/src/object_url.rs
hamza1311: I'm not entirely sure but I think `Rc::from` may be able to reuse the String heap allocation

WorldSEnder: I think that one still allocates twice, since the allocation inside an `Rc` contains also the two strong and weak counters, so can't just reuse the one from the `String`. Perhaps one could avoid the extra indirection?

The problem is this: `str` is unsized, so to get to an equivalent of `Rc<str>` and avoid the extra indirection, one would somehow have to unsize into it. On the other hand, the wrapper struct is important, as it provides the `Drop` impl that revokes the object url. Does a type that unsizes into `str` even exist? I'm not sure how I'd construct a struct containing a `str`, much less how to move that into an `Rc`.

hamza1311: Maybe we can remove the inner struct and move the Drop impl to this one?

I don't think we can avoid the extra indirection and heap allocation. I'm fine with keeping as it is

crates/file/src/object_url.rs
hamza1311: > For the most part, it's replicating `BlobContents`. The "overload" here is for the `Url::create_object_url` method.

I didn't know `BlobContents` also did it. I'd argue that should be changed as well but that's for another PR. From trait can also provide the overload and (imo) is better since it's part of std lib so users are more familiar with it and understand how that overload work. That may also help avoid looking at documentation to see how to construct it since the user can just use Intellisense with the From trait.

> A potential impl for `MediaSource` has been omitted because it's deprecated. Should it be added, too?

Unless someone complains and has a valid use-case, no, let's not add support deprecated APIs.

crates/file/src/object_url.rs
WorldSEnder: I at least changed the Deref target to `str` instead of `String`, that should make it easier to change this in the future.

hamza1311 (approved):

hamza1311:
Released in gloo-file v0.2.3 :tada:

PR #239, opened 2022-07-24
fix: Break dependency cycle by not using serde-serialize

andoriyu:
There is [an issue](https://github.com/tkaitchuck/aHash/issues/95) with cyclic dependency when `serde-serialize` feature is enabled in `wasm-bindgen`. Current suggestion is to use `js_sys` instead to avoid it.

I’ve checked and modified every crate that was enabling this feature:

 - `console` was easy to change
 - `storage` didn’t actually depend on this feature, so I just removed it.
 - `net` this one was a bit tricky and this the create that I couldn’t run tests for.

I ran non-wasm tests, and wasm tests for console and storage.

### DIFF
--- a/crates/console/Cargo.toml
+++ b/crates/console/Cargo.toml
@@ -12,9 +12,10 @@ documentation = "https://docs.rs/gloo-console/"
 categories = ["api-bindings", "development-tools::profiling", "wasm"]
 
 [dependencies]
-wasm-bindgen = { version = "0.2", features = ["serde-serialize"] }
+wasm-bindgen = "0.2"
 js-sys = "0.3"
 serde = { version = "1", features = ["derive"] }
+serde_json = "1.0"
 [dependencies.web-sys]
 version = "0.3"
 features = [
--- a/crates/console/src/lib.rs
+++ b/crates/console/src/lib.rs
@@ -38,7 +38,7 @@ pub mod __macro {
         data: impl serde::Serialize,
         columns: impl IntoIterator<Item = &'a str>,
     ) {
-        let data = JsValue::from_serde(&data).unwrap_throw();
+        let data = js_sys::JSON::parse(&serde_json::to_string(&data).unwrap_throw()).unwrap_throw();
         let columns = columns.into_iter().map(JsValue::from_str).collect();
 
         crate::externs::table_with_data_and_columns(data, columns);
--- a/crates/net/Cargo.toml
+++ b/crates/net/Cargo.toml
@@ -40,7 +40,7 @@ futures = "0.3"
 default = ["json", "websocket", "http"]
 
 # Enables `.json()` on `Response`
-json = ["wasm-bindgen/serde-serialize", "serde", "serde_json"]
+json = ["serde", "serde_json"]
 # Enables the WebSocket API
 websocket = [
     'web-sys/WebSocket',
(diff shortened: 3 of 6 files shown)


hamza1311:
Can you report the issue on wasm-bindgen repository? This PR seems like a bandaid over a much bigger issue. I'd much rather solve it at wasm-bindgen level and remove the cyclic dependency

andoriyu:
According to Alex that’s the correct way:
https://github.com/tkaitchuck/aHash/issues/95#issuecomment-881503355

There is no way to fix `wasm-bindgen` without splitting it into multiple
crates, there is already an issue for this:
https://github.com/rustwasm/wasm-bindgen/issues/2770

While it seems like a bandaid - many people can’t update their dependencies
because of this. Thread i linked earlier has references to crates that did
the same thing to break cycle.

On Sun, Jul 24, 2022 at 2:11 PM Muhammad Hamza ***@***.***>
wrote:Can you report the issue on wasm-bindgen repository? This PR seems
like a bandaid over a much bigger issue. I'd much rather solve it at
wasm-bindgen level and remove the cyclic dependency

hamza1311:
One way to fix it is to just remove the serde methods from wasm-bindgen. It'll be breaking change but it's one I'm willing to make. Gloo utils can provide a convert module with the 2 functions that are removed from wasm-bindgen 

@alexcrichton how does wasm-bindgen 0.3 sound?

andoriyu:
> One way to fix it is to just remove the serde methods from wasm-bindgen. It'll be breaking change but it's one I'm willing to make. Gloo utils can provide a convert module with the 2 functions that are removed from wasm-bindgen 
> 
> @alexcrichton how does wasm-bindgen 0.3 sound? 

I see. Yeah, that's reasonable. Is it possible to still merge this one (after everything passes) as a stop gap meanwhile?

alexcrichton:
Personally I don't think that `wasm-bindgen` 0.3 is the answer to this. As this shows taking the code from `wasm-bindgen` and putting it into an external crate is pretty easy and I think that's the better solution.

andoriyu:
@alexcrichton but if that feature stays in `wasm-bindgen`, then people keep ending up with that dependency cycle.

alexcrichton:
I am aware that is the consequence of inaction, yes.

ctron:
I would appreciate a fix on this too.

hamza1311:
Sorry, it took so long for me to get back to this. Things happened and I haven't had the time lately.

> I am aware that is the consequence of inaction, yes.

@alexcrichton, exactly. The dependency cycle will remain in wasm-bindgen. I'm not aware of any way to complete remove the cycle without removing the feature.

I'm personally in favor of 0.3. I'd like to go through and make other breaking changes that need to be made as part of wasm-bindgen 0.3. 

I'll merge this PR for now

> I can copy over those two functions to gloo-utils in this PR, standalone functions (`crate::json::{js_into_serde,js_from_serde}`) okay?

@andoriyu yeah, that would be good. I'm going to merge this once the CI is green so if you're up for this, please do so in a separate PR.

Also updating wasm-bindgen documentation to point to these functions would be a good idea. Until we remove the serde feature from wasm-bindgen, documenting the solution to the problem sounds like a good plan to me.

hamza1311:
The CI test failures are unrelated so I'll go ahead and merge this. Once again, apologies for the very late response.

PR #242, opened 2022-08-15
feat(gloo-utils): Lift serde-serialization from wasm-bindgen

andoriyu:
Per discussion in #239 adding support for `from_serde` and `into_serde` to `gloo-utils`.

I couldn't replicate functionality 1:1 because `wasm-bindgen` uses private API methods (`__wbindgen_json_serialize` and `__wbindgen_json_parse`), so I've reimplemented it with publicly available APIs.

`from_serde` stayed nearly the same, but throws an error in case `serde_json` gives an invalid JSON. It's probably safe to do `unwrap_unchecked`, but I wasn't sure which MSRV should I target.

`into_json` is more complicated. If you pass `undefined` to `JSON.stringify` it throws an error somewhere inside. Current behavior was to return an error, to keep current behavior, I've added a check if `JsValue` is `undefined`.

While at it, made changes to crates that were using `serde_json` (#239 PR) to use functions from `gloo-utils`. Test suite is lifted from `wasm-bindgen` nearly verbatim.

How `JSON.stringify` function behaves in a browser:
```
> JSON.stringify({w: undefined});
"{}"
> JSON.stringify(undefined);
undefined << causes panic in rust
```


andoriyu:
Fixed the formatting error. I thought I ran it, but I guess I didn't.

Clippy error & browser test failure looks like unrelated issue?

hamza1311:
Can you please merge master so CI is green?

crates/utils/Cargo.toml
```diff
@@ -27,5 +29,10 @@ features = [
     "Element",
 ]
 
+[features]
+default = ["serde-serialize"]
+serde-serialize = ["serde", "serde_json"]
```
hamza1311, line 34:
```suggestion
serde-serialize = ["dep:serde", "dep:serde_json"]
```

This makes sure that there's no feature called `serde` or `serde_json`. It does mean that MSRV for 1.61 (iirc) but 2 versions old at this point so I think it's fine, at least for this crate. For context, Yew also bumped the MSRV to 1.61 for this feature

I'd also rename the feature to `serde` since we deserialize as well

crates/utils/src/json.rs
```diff
@@ -0,0 +1,63 @@
...
+///
+/// # Errors
+///
+/// Returns any error encountered when serializing `T` into JSON.
+///
+/// # Panics
+///
+/// Panics if `serde_json` generated json that couldn't be parsed by `js_sys`.
+/// Uses `unwrap_throw` from `wasm_bindgen::UnwrapThrowExt`.
+#[cfg(feature = "serde-serialize")]
```
hamza1311, line 23: This needs `doc(cfg(_)` as well

crates/utils/src/json.rs
```diff
@@ -0,0 +1,63 @@
...
+///
+/// # Errors
+///
+/// Returns any error encountered when parsing the JSON into a `T`.
+///
+/// # Panics
+///
+/// Panics if `js_sys` couldn't stringify the JsValue. Uses `unwrap_throw`
+/// from `wasm_bindgen::UnwrapThrowExt`.
+#[cfg(feature = "serde-serialize")]
```
hamza1311, line 50: This needs `doc(cfg(_)` as well

crates/utils/src/json.rs
```diff
@@ -0,0 +1,63 @@
...
+/// the `gloo-utils` crate.
+///
+/// # Errors
+///
+/// Returns any error encountered when parsing the JSON into a `T`.
+///
+/// # Panics
+///
+/// Panics if `js_sys` couldn't stringify the JsValue. Uses `unwrap_throw`
+/// from `wasm_bindgen::UnwrapThrowExt`.
```
hamza1311, line 49:
```suggestion
/// Panics if [`js_sys`] couldn't stringify the JsValue. Uses [`unwrap_throw`](UnwrapThrowExt::unwrap_throw)
/// from [`wasm_bindgen::UnwrapThrowExt`].
```

This adds hyperlinks on docs.rs

crates/utils/src/json.rs
```diff
@@ -0,0 +1,63 @@
...
+/// the `gloo-utils` crate.
+///
+/// # Errors
+///
+/// Returns any error encountered when serializing `T` into JSON.
+///
+/// # Panics
+///
+/// Panics if `serde_json` generated json that couldn't be parsed by `js_sys`.
+/// Uses `unwrap_throw` from `wasm_bindgen::UnwrapThrowExt`.
```
hamza1311, line 22:
```suggestion
/// Panics if [`serde_json`](serde_json::to_string) generated JSON that couldn't be parsed by [`js_sys`].
/// Uses [`unwrap_throw`](UnwrapThrowExt::unwrap_throw) from [`wasm_bindgen::UnwrapThrowExt`].
```

crates/utils/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 pub mod errors;
 pub mod iter;
+pub mod json;
```
hamza1311, line 3:
```suggestion
pub mod format {
    pub mod json;
}
```

This makes it clear what `json` is from the module path. I think we can name `js_value_into_serde` and `js_value_from_serde` and reexport.

A better option is to have an extension trait that `JsValue` implements that provides these methods. That will also make the transition easier for wasm_bindgen users as all they need to do is add an import and remove the feature from Cargo.toml

crates/utils/tests/serde.rs
```diff
@@ -0,0 +1,59 @@
...
+    );
+    let foo = into_serde::<SerdeFoo>(&ret).unwrap();
+    assert_eq!(foo.a, 2);
+    assert_eq!(foo.b, "bar");
+    assert!(foo.c.is_some());
+    assert_eq!(foo.c.as_ref().unwrap().a, 3);
+    assert_eq!(foo.d.a, 4);
+    assert_eq!(into_serde::<String>(&JsValue::from("bar")).unwrap(), "bar");
+    assert_eq!(into_serde::<i32>(&JsValue::undefined()).ok(), None);
+    assert_eq!(into_serde::<i32>(&JsValue::null()).ok(), None);
```
hamza1311, line 58: Nit: maybe split into two tests, one for `from_serde`, one for `into_serde`

hamza1311 (changes requested):
We need to document the feature on docs.rs as well. This can done by:
- `lib.rs`
```rust
#![cfg_attr(docsrs, feature(doc_cfg))
```
- `on functions`
```rust
#[cfg_attr(docsrs, doc(cfg(feature = "_")))]
```
- `Cargo.toml`
```toml
[package.metadata.docs.rs]
all-features = true
rustdoc-args = ["--cfg", "docsrs"]
```

crates/utils/src/lib.rs
andoriyu: Got it. I will make it a trait instead.

andoriyu:
I think that covers all requested changes. I had to use UFCS in a few places because otherwise it confuses with `JsValue` methods since it's a trait now. 

Tests - the way I have written it, tests must be run in `node` environment rather than browser. I think otherwise, you can't use imports in modules?

andoriyu:
@hamza1311 do you want to keep:

```rust
let s = if self.is_undefined() {
    String::new()
} else {
    js_sys::JSON::stringify(self)
        .map(String::from)
        .unwrap_throw()
};
```

or do what `wasm-bindgen` does - wrap `JSON.stringify` in a function that replaces object with `null` if it's `undefined`?

There is also an option to use https://docs.rs/js-sys/latest/js_sys/JSON/fn.stringify_with_replacer.html, but I think it will slow down serialization too much?

crates/utils/src/lib.rs
```diff
@@ -1,5 +1,12 @@
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
 pub mod errors;
 pub mod iter;
+pub mod format {
+    mod json;
```
Liamolucko, line 6: The fact that this uses JSON feels like an implementation detail to me; I think `serde` would be a better name.

hamza1311: I agree that serde is a better. But it's not exposed, so implementation detail is only visible if inspecting the source code, as it should be, so I think it's fine

crates/utils/src/format/json.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use wasm_bindgen::{JsValue, UnwrapThrowExt};
+mod private {
+    pub trait Sealed {}
+    impl Sealed for wasm_bindgen::JsValue {}
+}
+
+/// Extenstion trait to provide conversion between [`JsValue`](wasm_bindgen::JsValue) and [`serde`].
+///
+/// Usage of this API requires activating the `serde` feature of the `gloo-utils` crate.
```
hamza1311, line 11: An example here would be nice to have

crates/net/src/websocket/futures.rs
```diff
@@ -106,16 +107,12 @@ impl WebSocket {
         url: &str,
         protocols: &[S],
     ) -> Result<Self, JsError> {
-        let map_err = |err| {
+        let json = <JsValue as JsValueSerdeExt>::from_serde(protocols).map_err(|err| {
```
hamza1311, line 110: Does `JsValue::from_serde` not work?

crates/net/src/websocket/futures.rs
```diff
@@ -33,6 +33,7 @@ use futures_channel::mpsc;
 use futures_core::{ready, Stream};
 use futures_sink::Sink;
 use gloo_utils::errors::JsError;
+use gloo_utils::format::JsValueSerdeExt;
```
hamza1311, line 36: Can you remove this if it's not needed?

crates/utils/tests/serde.js
```diff
@@ -0,0 +1,20 @@
+const wasm = require('wasm-bindgen-test.js');
```
hamza1311, line 1: This is unused so it can be removed

crates/utils/tests/serde.js
```diff
@@ -0,0 +1,20 @@
...
+const assert = require('assert');
```
hamza1311, line 2: Only `deepStrictEquals` is used from this module, which can be implemented as

```js
function deepStrictEquals(left, right) {
    if (left !== right) {
        throw Error(`${left} != ${right}`}
    }
}
```
That way, we can run these tests in browser too

hamza1311 (commented):
> I'm not sure why unrelated test failed.

The tests are all green on master now (as of #244 thanks to @futursolo). If you merge master here, the only failing tests should be from this PR

crates/utils/tests/serde.js
andoriyu: oh, thanks. I will change it right now.

crates/net/src/websocket/futures.rs
andoriyu: @hamza1311 The issue with `JsValue::from_serde` - `JsValue` currently has a method with the same name if you have `serde-serialization` feature enabled. I wanted to avoid confusion which method is used.

Even when the feature is disabled, `cargo check` was complaining that import isn't used even though it's used.

crates/utils/src/format/json.rs
```diff
@@ -0,0 +1,86 @@
...
+        Ok(js_sys::JSON::parse(&s).unwrap_throw())
+    }
+
+    fn into_serde<T>(&self) -> serde_json::Result<T>
+    where
+        T: for<'a> serde::de::Deserialize<'a>,
+    {
+        let s = if self.is_undefined() {
+            String::new()
+        } else {
```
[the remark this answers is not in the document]
andoriyu: The issue is actually here: https://github.com/rustwasm/gloo/pull/242/files/7aa660a8ba3213aca2cb3bc302a2c12b775b386a#diff-c6ef873253c87613c2afa2ac17e7ffae4074106c602e6c15d7aa6dc23df12bbaR80

`JSON.stringify` panics when you give it a `JsValue` that is `undefined`. I can do what `wasm-bindgen` does it, replace `undefined` with `null`.

crates/net/src/websocket/futures.rs
andoriyu: It's used on line 110 though.

crates/utils/tests/serde.js
andoriyu: Turns you it's a little more involved than that, but this use case I have a work around.

crates/utils/src/lib.rs
Liamolucko: Oops, sorry, didn't notice that the `mod json` was private.

crates/utils/src/format/json.rs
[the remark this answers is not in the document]
Liamolucko: You could also return the string `"null"` directly, since that's what `JSON.stringify` will always produce when passed `null`.

crates/net/src/websocket/futures.rs
andoriyu: Ah, it's unused unless som feature is enabled. Fixed.

crates/utils/src/format/json.rs
andoriyu: Done

crates/utils/src/format/json.rs
```diff
@@ -0,0 +1,114 @@
...
+mod private {
+    pub trait Sealed {}
+    impl Sealed for wasm_bindgen::JsValue {}
+}
+
+// Turns out `JSON.stringify(undefined) === undefined`, so if
+// we're passed `undefined` reinterpret it as `null` for JSON
+// purposes.
+#[wasm_bindgen(
+    inline_js = "export function serialize(obj) { return JSON.stringify(obj === undefined ? null : obj) }"
```
futursolo, line 14: Inline JS snippets could cause issues for non-ES module targets.
`nomodule` target is still the only target supported by `gloo-worker`.

See: https://rustwasm.github.io/docs/wasm-bindgen/examples/without-a-bundler.html#using-the-older---target-no-modules

Instead of going through `JSON::stringify` and `serde_json`, I would suggest using [`serde-wasm-bindgen`](https://docs.rs/serde-wasm-bindgen/latest/serde_wasm_bindgen/). It supports non-serializable types (e.g.: ES2015 `Map`) and has a smaller bundle footprint than `serde_json`.

andoriyu: @futursolo oh well, back to what I had before.

So if you use [`serde_wasm_bindgen`](https://docs.rs/serde-wasm-bindgen/latest/serde_wasm_bindgen/), then you don't really need this. The goal is to provide a drop-in replacement for `JsValue::{from_serde/to_serde}` and an easier way to use `js_sys::JSON::`.

Plus, in the deprecation thread, [someone](https://github.com/rustwasm/wasm-bindgen/pull/3031#issuecomment-1216561512) said they had a bad time with `serde-wasm-bindgen`. I think it makes sense: it calls to JS for every field which is slow.

hamza1311 (approved):
Looks good to me. Thanks a lot for your work!

hamza1311:
These changes have been released in:

**gloo-utils**: v0.1.5
**gloo-net**: v0.2.4
**gloo-storage**: v0.2.2
**gloo-console**: v0.2.3

PR #244, opened 2022-08-17
Fix gloo-worker tests

futursolo:
This pull request fixes an issue where: 
- gloo-worker tests will fail if test server does not start up before tests are run.
- wasm-bindgen now requires CORP headers.


hamza1311 (approved):
Thanks for the fix :+1:

Issue #8, opened 2019-02-26
Server sent events library


fitzgen:
Perhaps something like https://github.com/yerkopalma/choo-sse?

Probably best expressed as a futures stream.

dureuill:
Hello!

I have been following the Server-sent-events (SSE) libraries for yew closely, and I even built an [example](https://github.com/dureuill/yew-sse-example) of using [liquidnya/yew-sse](https://github.com/liquidnya/yew-sse) with yew 16.2, rocket 0.5-rc1 and trunk.

If I understand correctly, yew::services are being removed so that the yew crate can be thinner which will ease server-side-rendering (SSR)?

I'm unfamiliar with gloo, but willing to help to get SSE into yew. Is the correct way forward to open a PR against gloo? What about reqwasm that is mentioned in https://github.com/yewstack/yew/issues/1346#issuecomment-883469926? I'm a bit unfamiliar with the WASM ecosystem.

PR #246, opened 2022-08-23
Provides an EventSource implementation

huntc:
Inspired by the WebSocket API.

Example usage:

```rust
use gloo_net::eventsource::futures::EventSource;
use wasm_bindgen_futures::spawn_local;
use futures::{stream, StreamExt};

let mut es = EventSource::new("http://api.example.com/ssedemo.php").unwrap();
let stream_1 = es.subscribe("some-event-type").unwrap();
let stream_2 = es.subscribe("another-event-type").unwrap();

spawn_local(async move {
    let mut all_streams = stream::select(stream_1, stream_2);
    while let Some(Ok((event_type, msg))) = all_streams.next().await {
        console_log!(format!("1. {}: {:?}", event_type, msg))
    }
    console_log!("EventSource Closed");
})
```

TODO:

* [x] Find/build an EventSource endpoint for testing
* [x] Test from within my real-world application

Fixes https://github.com/hamza1311/reqwasm/issues/8
Fixes https://github.com/rustwasm/gloo/issues/8

### DIFF
--- a/crates/net/Cargo.toml
+++ b/crates/net/Cargo.toml
@@ -19,7 +19,6 @@ wasm-bindgen = "0.2"
 web-sys = "0.3"
 js-sys = "0.3"
 gloo-utils = { version = "0.1", path = "../utils", features = ["serde"] }
-gloo-console = { path = "../console" }
 
 wasm-bindgen-futures = "0.4"
 futures-core = { version = "0.3", optional = true }
--- a/crates/net/Cargo.toml
+++ b/crates/net/Cargo.toml
@@ -19,6 +19,7 @@ wasm-bindgen = "0.2"
 web-sys = "0.3"
 js-sys = "0.3"
 gloo-utils = { version = "0.1", path = "../utils", features = ["serde"] }
+gloo-console = { path = "../console" }
 
 wasm-bindgen-futures = "0.4"
 futures-core = { version = "0.3", optional = true }
@@ -37,7 +38,7 @@ wasm-bindgen-test = "0.3"
 futures = "0.3"
 
 [features]
-default = ["json", "websocket", "http"]
+default = ["json", "websocket", "http", "eventsource"]
 
 # Enables `.json()` on `Response`
 json = ["serde", "serde_json"]
@@ -78,4 +79,14 @@ http = [
     'web-sys/ReadableStream',
     'web-sys/Blob',
     'web-sys/FormData',
+]
+# Enables the EventSource API
+eventsource = [
+    "futures-channel",
+    "futures-core",
+    "pin-project",
+    'web-sys/Event',
+    'web-sys/EventTarget',
+    'web-sys/EventSource',
+    'web-sys/MessageEvent',
 ]
--- a/crates/net/README.md
+++ b/crates/net/README.md
@@ -19,7 +19,7 @@
 <sub>Built with 🦀🕸 by <a href="https://rustwasm.github.io/">The Rust and WebAssembly Working Group</a></sub>
 </div>
 
-HTTP requests library for WASM Apps. It provides idiomatic Rust bindings for the `web_sys` `fetch`, `WebSocket` and `EventSource` APIs.
+HTTP requests library for WASM Apps. It provides idiomatic Rust bindings for the `web_sys` [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), [`WebSocket`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket) and [`EventSource`](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) APIs.
 
 ## Examples
 
--- a/crates/net/README.md
+++ b/crates/net/README.md
@@ -19,7 +19,7 @@
 <sub>Built with 🦀🕸 by <a href="https://rustwasm.github.io/">The Rust and WebAssembly Working Group</a></sub>
 </div>
 
-HTTP requests library for WASM Apps. It provides idiomatic Rust bindings for the `web_sys` `EventSource`, `fetch` and `WebSocket` APIs.
+HTTP requests library for WASM Apps. It provides idiomatic Rust bindings for the `web_sys` `fetch`, `WebSocket` and `EventSource` APIs.
 
 ## Examples
 
@@ -64,8 +64,8 @@ use wasm_bindgen_futures::spawn_local;
 use futures::StreamExt;
 
 let mut es = EventSource::new("http://api.example.com/ssedemo.php").unwrap();
-es.subscribe_event("some-event-type").unwrap();
-es.subscribe_event("another-event-type").unwrap();
+es.subscribe("some-event-type").unwrap();
+es.subscribe("another-event-type").unwrap();
 
 spawn_local(async move {
     while let Some(Ok((event_type, msg))) = es.next().await {
--- a/crates/net/README.md
+++ b/crates/net/README.md
@@ -61,14 +61,15 @@ spawn_local(async move {
 ```rust
 use gloo_net::eventsource::futures::EventSource;
 use wasm_bindgen_futures::spawn_local;
-use futures::StreamExt;
+use futures::{stream, StreamExt};
 
 let mut es = EventSource::new("http://api.example.com/ssedemo.php").unwrap();
-es.subscribe("some-event-type").unwrap();
-es.subscribe("another-event-type").unwrap();
+let stream_1 = es.subscribe("some-event-type").unwrap();
+let stream_2 = es.subscribe("another-event-type").unwrap();
 
 spawn_local(async move {
-    while let Some(Ok((event_type, msg))) = es.next().await {
+    let mut all_streams = stream::select(stream_1, stream_2);
+    while let Some(Ok((event_type, msg))) = all_streams.next().await {
         console_log!(format!("1. {}: {:?}", event_type, msg))
     }
     console_log!("EventSource Closed");
--- a/crates/net/README.md
+++ b/crates/net/README.md
@@ -19,7 +19,7 @@
 <sub>Built with 🦀🕸 by <a href="https://rustwasm.github.io/">The Rust and WebAssembly Working Group</a></sub>
 </div>
 
-HTTP requests library for WASM Apps. It provides idiomatic Rust bindings for the `web_sys` `fetch` and `WebSocket` API
+HTTP requests library for WASM Apps. It provides idiomatic Rust bindings for the `web_sys` `EventSource`, `fetch` and `WebSocket` APIs.
 
 ## Examples
 
@@ -53,5 +53,24 @@ spawn_local(async move {
         console_log!(format!("1. {:?}", msg))
     }
     console_log!("WebSocket Closed")
+})
+```
+
+### EventSource
+
+```rust
+use gloo_net::eventsource::futures::EventSource;
+use wasm_bindgen_futures::spawn_local;
+use futures::StreamExt;
+
+let mut es = EventSource::new("http://api.example.com/ssedemo.php").unwrap();
+es.subscribe_event("some-event-type").unwrap();
+es.subscribe_event("another-event-type").unwrap();
+
+spawn_local(async move {
+    while let Some(Ok((event_type, msg))) = es.next().await {
+        console_log!(format!("1. {}: {:?}", event_type, msg))
+    }
+    console_log!("EventSource Closed");
 })
 ```
--- a/crates/net/src/eventsource/futures.rs
+++ b/crates/net/src/eventsource/futures.rs
@@ -101,7 +101,7 @@ impl EventSource {
         })
     }
 
-    /// Subscribes to listening for a specific type of event. Can be
+    /// Subscribes to listening for a specific type of event. This method can be
     /// called multiple times. Subscribing again with an event type
     /// that has already been subscribed is benign.
     ///
--- a/crates/net/src/eventsource/futures.rs
+++ b/crates/net/src/eventsource/futures.rs
@@ -15,18 +15,19 @@
 //! ```rust
 //! use gloo_net::eventsource::futures::EventSource;
 //! use wasm_bindgen_futures::spawn_local;
-//! use futures::StreamExt;
+//! use futures::{stream, StreamExt};
 //!
 //! # macro_rules! console_log {
 //! #    ($($expr:expr),*) => {{}};
 //! # }
 //! # fn no_run() {
 //! let mut es = EventSource::new("http://api.example.com/ssedemo.php").unwrap();
-//! es.subscribe("some-event-type").unwrap();
-//! es.subscribe("another-event-type").unwrap();
+//! let stream_1 = es.subscribe("some-event-type").unwrap();
+//! let stream_2 = es.subscribe("another-event-type").unwrap();
 //!
 //! spawn_local(async move {
-//!     while let Some(Ok((event_type, msg))) = es.next().await {
+//!     let mut all_streams = stream::select(stream_1, stream_2);
+//!     while let Some(Ok((event_type, msg))) = all_streams.next().await {
 //!         console_log!(format!("1. {}: {:?}", event_type, msg))
 //!     }
 //!     console_log!("EventSource Closed");
@@ -39,8 +40,6 @@ use futures_channel::mpsc;
 use futures_core::{ready, Stream};
 use gloo_utils::errors::JsError;
 use pin_project::{pin_project, pinned_drop};
-use std::collections::hash_map::Entry;
-use std::collections::HashMap;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 use wasm_bindgen::prelude::*;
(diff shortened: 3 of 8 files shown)


crates/net/Cargo.toml
```diff
@@ -37,7 +38,7 @@ wasm-bindgen-test = "0.3"
 futures = "0.3"
 
 [features]
-default = ["json", "websocket", "http"]
+default = ["json", "websocket", "http", "eventsource"]
```
hseeberger, line 41: Generally I am reluctant to putting "too" many features into the default one, but as we already have websocket here I guess we can also have eventsource.

crates/net/README.md
```diff
@@ -19,7 +19,7 @@
 <sub>Built with 🦀🕸 by <a href="https://rustwasm.github.io/">The Rust and WebAssembly Working Group</a></sub>
 </div>
 
-HTTP requests library for WASM Apps. It provides idiomatic Rust bindings for the `web_sys` `fetch` and `WebSocket` API
+HTTP requests library for WASM Apps. It provides idiomatic Rust bindings for the `web_sys` `EventSource`, `fetch` and `WebSocket` APIs.
```
hseeberger, line 22: Maybe reorder "fetch , EventSource" to sort by specificy/power.

crates/net/src/eventsource/futures.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// that has already been subscribed is benign.
+    ///
+    /// All event types are streamed back with the element of the stream
+    /// being a tuple of event type and message event.
+    ///
+    /// The event type of "message" is a special case, as it will capture
+    /// events without an event field as well as events that have the
+    /// specific type `event: message`. It will not trigger on any
+    /// other event type.
+    pub fn subscribe_event(&mut self, event_type: &str) -> Result<(), JsError> {
```
hseeberger, line 119: I suggest to rename this function to just `subscribe`. And the same for `unsubscribe` below. It's obvious that we are dealing with events and there is nothing else to subsrcibe to.

hseeberger (approved):
For the maintainers of this repository: I am the original author of SSE support in Akka HTTP (see https://github.com/hseeberger/akka-sse) and have been working with Rust in production for some years, also created some OSS libraries (e.g. https://crates.io/crates/pub-sub-client).

Thanks Chris for this excellent PR. I only have some minor suggestions and I'd love to see this being merged.

crates/net/README.md
hamza1311:
```suggestion
HTTP requests library for WASM Apps. It provides idiomatic Rust bindings for the `web_sys` `EventSource` (SSE), `fetch` and `WebSocket` APIs.
```

Also, it would be nice to add MDN links for `EventSource`, `fetch` and `WebSocket`

crates/net/src/eventsource/futures.rs
```diff
@@ -0,0 +1,258 @@
...
+#[allow(missing_debug_implementations)]
+#[pin_project(PinnedDrop)]
+pub struct EventSource {
+    es: web_sys::EventSource,
+    message_sender: mpsc::UnboundedSender<StreamMessage>,
+    #[pin]
+    message_receiver: mpsc::UnboundedReceiver<StreamMessage>,
+    #[allow(clippy::type_complexity)]
+    closures: Arc<
+        Mutex<(
```
hamza1311, line 62: Why do we need interior mutability? All the methods that mutate take `&mut self` and we don't implement `Clone` so there's no need got `Arc<Mutex<_>>`

crates/net/src/eventsource/futures.rs
```diff
@@ -0,0 +1,258 @@
...
+        Ok(Self {
+            es,
+            message_sender,
+            message_receiver,
+            closures: Arc::new(Mutex::new((HashMap::new(), error_callback))),
+        })
+    }
+
+    /// Subscribes to listening for a specific type of event. Can be
+    /// called multiple times. Subscribing again with an event type
```
hamza1311, line 109:
```suggestion
    /// Subscribes to listening for a specific type of event. This method can be
    /// called multiple times. Subscribing again with an event type
```

crates/net/src/eventsource/futures.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        spawn_local(async move {
+            assert_eq!(es.next().await.unwrap().unwrap().0, "server");
+            assert_eq!(es.next().await.unwrap().unwrap().0, "request");
+            es.unsubscribe_event("request").unwrap();
+        });
+    }
+
+    #[wasm_bindgen_test]
+    fn eventsource_close_works() {
```
hamza1311, line 246:
```suggestion
    fn eventsource_connection_failure_works() {
```

Testing close will mean `EventSource::close` is has test. Can you add a test for that?

crates/net/src/eventsource/futures.rs
```diff
@@ -0,0 +1,258 @@
...
+impl EventSource {
+    /// Establish an EventSource.
+    ///
+    /// This function may error in the following cases:
+    /// - The connection url is invalid
+    ///
+    /// The error returned is [`JsError`]. See the
+    /// [MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/EventSource/EventSource#exceptions_thrown)
+    /// to learn more.
+    pub fn new(url: &str) -> Result<Self, JsError> {
```
hamza1311, line 78: I think we should also have a function `new_with_subscriptions` which calls new and subscribes to the events passed in

It can made more efficient as there's no need to create an empty HashMap and then mutate it later on.

crates/net/Cargo.toml
huntc: Yeah, seemed to make sense to continue following the convention.

crates/net/README.md
huntc: When in doubt, I put things in alphabetical order. Happy to reorder here though.

crates/net/src/eventsource/futures.rs
```diff
@@ -0,0 +1,258 @@
...
+
+        let (message_sender, message_receiver) = mpsc::unbounded();
+
+        let error_callback: Closure<dyn FnMut(web_sys::Event)> = {
+            let sender = message_sender.clone();
+            Closure::wrap(Box::new(move |e: web_sys::Event| {
+                let sender = sender.clone();
+                let is_connecting = e
+                    .current_target()
+                    .and_then(|target| target.dyn_into::<web_sys::EventSource>().ok())
```
hamza1311, line 89:
```suggestion
                    .map(|target| target.unchecked_into::<web_sys::EventSource>())
```
Any reason not to use `unchecked_into`? If we're sure about the type, why perform a check?

crates/net/src/eventsource/futures.rs
huntc: Will do

crates/net/src/eventsource/futures.rs
```diff
@@ -0,0 +1,258 @@
...
+        es.set_onerror(Some(error_callback.as_ref().unchecked_ref()));
+
+        Ok(Self {
+            es,
+            message_sender,
+            message_receiver,
+            closures: Arc::new(Mutex::new((HashMap::new(), error_callback))),
+        })
+    }
+
```
hamza1311, line 107:
```suggestion
    pub fn subscribe_all(&mut self) -> Result<(), JsError> { ... }
```
This will listen to all events emitted (`onmessage`)

crates/net/src/eventsource/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+    /// The EventSource connection is established and communication is possible.
+    Open,
+    /// The connection has been closed or could not be opened.
+    Closed,
+}
+
+/// Error returned by the EventSource
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum EventSourceError {
```
hamza1311, line 27: Why not use `thiserror` here?

huntc:
Commit 1ea4f0a39680c92d6d4fb77be2bd6baeff54c3ef addresses @hseeberger's feedback. Thanks for the review, Heiko!

crates/net/src/eventsource/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+    /// The connection has not yet been established.
+    Connecting,
+    /// The EventSource connection is established and communication is possible.
+    Open,
+    /// The connection has been closed or could not be opened.
+    Closed,
+}
+
+/// Error returned by the EventSource
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```
hamza1311, line 25:
```suggestion
#[derive(Clone, Debug, Eq, PartialEq)]
```
`Copy` means we can't add a error message returned from the `EventSource` in the future

huntc:
Unsure why the browser tests are failing on CI. Doesn't appear related.

hamza1311 (changes requested):
Why not use [`gloo_events::EventListener`](https://docs.rs/gloo-events/latest/gloo_events/struct.EventListener.html) instead of manually registering callbacks? 

Also. I'm not sure if this is the best way to expose the API. What if someone wants to subscribe to different events at different places? I would prefer if each subscribe method would return a `EventSubscription` that can be consumed independently of the `EventSource`. Of course, this causes complications with closing the underlying `web_sys::EventSource`. One solution I can think of is to close the connection when all the `EventSubscription`s are dropped. A method to forcefully close the connection can also be provided but with a warning that there may be `EventSubscription`s that are still using this connection. Waiting on them will lead to a deadlock

```rust
pub struct EventSource { ... }

impl EventSource {
    fn new(...) -> Self;
    fn subscribe_event(&self, event: &str) -> EventSubscription;
    fn subscribe_all(&self) -> EventSubscription { self.subscribe_event("message") }
}

struct EventSubscription { ... }
impl EventSubscription {
    // This is our event source. It can Rc'd to be passed around
    fn event_source(&self) -> &EventSource { ... }
}
impl Stream for EventSubscription { ... }
impl Drop for EventSubscription { unsubscribe the event }

**Comment by huntc** [comment]
> Why not use [`gloo_events::EventListener`](https://docs.rs/gloo-events/latest/gloo_events/struct.EventListener.html) instead of manually registering callbacks?

I'm unsure what the stringified event names should be when registering in a more generic way. I think it might be `message`, but then I'm unsure how the event type to subscribe to would be used.

How we actually register the event types is an implementation detail though. Would you mind if we defer to an improved method in a subsequent PR if you feel it would benefit the code?

> 
> Also. I'm not sure if this is the best way to expose the API. What if someone wants to subscribe to different events at different places? I would prefer if each subscribe method would return a `EventSubscription` that can be consumed independently of the `EventSource`. Of course, this causes complications with closing the underlying `web_sys::EventSource`. One solution I can think of is to close the connection when all the `EventSubscription`s are dropped. A method to forcefully close the connection can also be provided but with a warning that there may be `EventSubscription`s that are still using this connection. Waiting on them will lead to a deadlock
> 
> ```rust
> pub struct EventSource { ... }
> 
> impl EventSource {
>     fn new(...) -> Self;
>     fn subscribe_event(&self, event: &str) -> EventSubscription;
>     fn subscribe_all(&self) -> EventSubscription { self.subscribe_event("message") }
> }
> 
> struct EventSubscription { ... }
> impl EventSubscription {
>     // This is our event source. It can Rc'd to be passed around
>     fn event_source(&self) -> &EventSource { ... }
> }
> impl Stream for EventSubscription { ... }
> impl Drop for EventSubscription { unsubscribe the event }
> ```

I think having a stream per subscription adds a bit of complexity, so I'm not sure it is a great idea. In my experience with SSE, I've found that I typically consume all events in one place. However, I appreciate that what you're suggesting is perhaps more in line with the underlying API. Do we really want multiple streams being driven and consumed though? Again, I'm not convinced that the complexity is worth it.

Also, I don't believe there's a way of subscribing to all events. There's an on_message event handler, but I believe that it only fires when there is no event type specified in the source data. If you subscribe to `message` as an event type then it also yields those events with no event type. Again though, I don't believe you can subscribe to all events.

File: crates/net/README.md (hunk 2 above)
Comment (huntc) [reply crates/net/README.md]: Commit 2490c1ebe89440d0fc2748b330fc10eb3cce5ce3 addresses this.

File: crates/net/src/eventsource/futures.rs (hunk 4 above)
Comment (huntc) [reply crates/net/src/eventsource/futures.rs:62]: Great catch! Dunno what I was thinking there.

Commit c23e56942b38d2248cfb80e15009bdaf2d1966f0 addresses this.

File: crates/net/src/eventsource/futures.rs (hunk 5 above)
Comment (huntc) [reply crates/net/src/eventsource/futures.rs:109]: Commit f815888847dce394ad86decce0c9bd8324f703e3 addresses this.

File: crates/net/src/eventsource/futures.rs (hunk 6 above)
Comment (huntc) [reply crates/net/src/eventsource/futures.rs:246]: This test was named incorrectly. I've now renamed it to `eventsource_connect_failure_works`.

Commit 4688b89ac79c24124fef4a9460c9761509792100 addresses this.

File: crates/net/src/eventsource/futures.rs (hunk 7 above)
Comment (huntc) [reply crates/net/src/eventsource/futures.rs:78]: I'm unsure about this as we'd still be mutating a hashmap in the constructor... The downside of doing this would also be more code in the implementation.

File: crates/net/src/eventsource/futures.rs (hunk 8 above)
Comment (huntc) [reply crates/net/src/eventsource/futures.rs:89]: Good catch. Commit c2434b4ea216c8c8ac27e7ded324fe73ef3cdd0e addresses this.

File: crates/net/src/eventsource/futures.rs (hunk 9 above)
Comment (huntc) [reply crates/net/src/eventsource/futures.rs:107]: I do not believe this can be implemented. I believe that registering the `on_message` handler will return just those events with no event type.

File: crates/net/src/eventsource/mod.rs (hunk 10 above)
Comment (huntc) [reply crates/net/src/eventsource/mod.rs:27]: I'm unfamiliar with `thiserror`. However, am I not doing the same thing here as per `WebSocket` by way of declaring the error type?

File: crates/net/src/eventsource/mod.rs (hunk 11 above)
Comment (huntc) [reply crates/net/src/eventsource/mod.rs:25]: This `lib.rs` file was causing this to occur I think:

```rust
#![deny(
    missing_docs,
    missing_debug_implementations,
    missing_copy_implementations
)]
```

Commit 6d6cf4f69e614be46073759182e531f88d6ca0de addreses this by permitting it for the error type.

huntc:
> I think having a stream per subscription adds a bit of complexity, so I'm not sure it is a great idea. In my experience with SSE, I've found that I typically consume all events in one place. However, I appreciate that what you're suggesting is perhaps more in line with the underlying API. Do we really want multiple streams being driven and consumed though? Again, I'm not convinced that the complexity is worth it.

I had to scratch this itch to see how it looks. I'm actually quite happy with it and have now pushed a commit that implements what you suggested. The commit is fa096f3fe9a639d0b600d1b83714774303351d16.

huntc:
I'm going to revert this back to a draft for now while I get the subscription stream approach working. I'm really unsure about the tests running properly also, so I want to delve into that a bit more.

huntc:
OK, this is ready for review again. I've checked that it continues to work given another app. I'm unhappy that I'm not able to get the tests running locally - well, they run, but I can't seem to get them to fail. I'm also not convinced that the CI issue being reported is related.

Other than that, I'm happy with the approach taken here given the fabulous feedback received.

crates/net/src/eventsource/futures.rs
hseeberger: Not sure if this efficiency gain is worth the additional complexity. Personally I am happy with creating an EventSource and then subscribing to it.

crates/net/src/eventsource/futures.rs
hseeberger: @huntc is right, according to the spec registering the `on_message` handler only listens to unnamed events, i.e. these without type.

crates/net/src/eventsource/futures.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// Subscribes to listening for a specific type of event.
+    ///
+    /// All events for this type are streamed back given the subscription
+    /// returned.
+    ///
+    /// The event type of "message" is a special case, as it will capture
+    /// events without an event field as well as events that have the
+    /// specific type `event: message`. It will not trigger on any
+    /// other event type.
+    pub fn subscribe(&mut self, event_type: &str) -> Result<EventSourceSubscription, JsError> {
```
hseeberger, line 94: I really like this change (returning a subscription). Nicer API and nicer implementation. Great feedback from @hamza1311 .

crates/net/src/eventsource/mod.rs
hseeberger: I really like `thiserror`, but I agree with @huntc that this crate (at least in websocket/mod.rs) is not using `thiserror`. Is this something we should consider changing consistently in another PR, @hamza1311? I'd be glad to provide one.

crates/net/src/eventsource/futures.rs
hamza1311: There's `HashMap::with_capacity` but it's fine to leave it as it is as well

crates/net/src/eventsource/futures.rs
```diff
@@ -0,0 +1,252 @@
...
+/// Wrapper around browser's EventSource API. Dropping
+/// this will close the underlying event source.
+#[allow(missing_debug_implementations)]
+#[derive(Clone)]
+pub struct EventSource {
+    es: web_sys::EventSource,
+}
+
+/// Wrapper around browser's EventSource API.
+#[allow(missing_debug_implementations)]
```
hamza1311, line 58: A manual `Debug` implementation that gives all the useful information we can extract from this type would be helpful.

crates/net/src/eventsource/futures.rs
```diff
@@ -0,0 +1,252 @@
...
+    ///
+    /// All events for this type are streamed back given the subscription
+    /// returned.
+    ///
+    /// The event type of "message" is a special case, as it will capture
+    /// events without an event field as well as events that have the
+    /// specific type `event: message`. It will not trigger on any
+    /// other event type.
+    pub fn subscribe(&mut self, event_type: &str) -> Result<EventSourceSubscription, JsError> {
+        let event_type = event_type.to_string();
```
hamza1311, line 95: This seems like an unnecessary clone. What happens if you remove it?

crates/net/src/eventsource/mod.rs
hamza1311: @hseeberger if you want to PR converting this to `thiserror`, that would be great.

It's fine as is for this PR

hamza1311 (commented):
Sorry, it took me so long to review this. I've been busy with work and real life stuff.

I'll address the review comments myself and merge it

hseeberger:
Will do thiserror eventually …

hamza1311 (approved):

huntc:
> Sorry, it took me so long to review this. I've been busy with work and real life stuff.
> 

No problem at all! Thanks  to you and @hseeberger for the excellent reviews. Hope real life stuff is ok with you.

PR #249, opened 2022-08-29
Work around firefox webextensions bug

DCNick3:
Importing `gloo_timers` in firefox webextension's content script results in an error: `LinkError: import object field '__wbg_clearTimeout_65417660fe82f08d' is not a Function`

It actually seems to be a firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1787770

This can simply be worked around by refering to timer functions using window object (setInterval -> window.setInterval) which this PR does. I do not think this would impact users outside of webextensions in any way, as window.setInterval and setInterval are the same functions in usual context (but apparently not in content scripts...)

### DIFF
--- a/crates/timers/src/callback.rs
+++ b/crates/timers/src/callback.rs
@@ -6,16 +6,16 @@ use wasm_bindgen::{JsCast, JsValue};
 
 #[wasm_bindgen]
 extern "C" {
-    #[wasm_bindgen(js_name = "setTimeout", catch)]
+    #[wasm_bindgen(js_name = "setTimeout", js_namespace = window, catch)]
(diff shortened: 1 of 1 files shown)


DCNick3:
Well, looks like this work around is not good enough. While the wasm module does instantiate fine, it still raises an error `TypeError: 'clearTimeout' called on an object that does not implement interface Window.` when the function is actually invoked.

DCNick3:
Now this should work. Instead of using `window` object to refer to the function I make wasm-bindgen generate a wrapper around the desired function by marking it with `catch`. This way wasm module does not error out on instantiation and is actually able to use those APIs

hamza1311:
It seems as a fix for the Firefox bug is underway. I would say we shouldn't merge this and let Firefox handle it.

Issue #201, opened 2022-03-24
gloo-net 


cdata:
### Describe the Bug

`gloo-net` HTTP features cannot be used in a worker because the library always tries to access the Fetch API from `Window`, which will fail when running in a worker: https://github.com/rustwasm/gloo/blob/5fc0a8ffbed68c9a5ce76e9a90bafa204720717a/crates/net/src/http/mod.rs#L189

### Steps to Reproduce

1. Design a WASM-compatible program that uses `gloo-net` to make HTTP requests
2. Load and run it in a Web Worker

### Expected Behavior

Expected HTTP requests are made.

### Actual Behavior

An error, as the library attempts to access the wrong kind of global.

### Additional Context

In digging around, I noticed there is some relevant history from `gloo` contributors: https://github.com/rustwasm/wasm-bindgen/issues/2148#issuecomment-638606446.

Following that comment thread, I discovered that gloo previously incorporated a limited workaround for accessing global properties in a scope-agnostic way: https://github.com/rustwasm/gloo/blob/2c9e776701ecb90c53e62dec1abd19c2b70e47c7/crates/timers/src/callback.rs#L8-L40

In some cases, it may feasible to take a simpler approach, such as this patch I applied to another crate that has this same issue: https://github.com/devashishdxt/rexie/commit/8637e4ebc2d2dfc6b33cd60dc85b99e46d6afa96

cdata:
I have created a fairly ugly patch to make `gloo-net` usable in `DedicatedWorkerGlobalScope` contexts: https://github.com/rustwasm/gloo/compare/master...cdata:master

allsey87:
> I have created a fairly ugly patch to make `gloo-net` usable in `DedicatedWorkerGlobalScope` contexts: [master...cdata:master](https://github.com/rustwasm/gloo/compare/master...cdata:master)

I wonder if a better solution would be to solve this at the wasm-bindgen level with a `Global` trait that is returned by `js_sys::global` and that covers all of the common APIs between `Window` and `DedicatedWorkerGlobalScope`.

Perhaps @alexcrichton has some thoughts on that?

Zageron:
@cdata Thank you very much for your fork. I assume it's quite out of date now, but it still managed to resolve my issue.

cdata:
@Zageron I don't have time to look at it right now, but if you happen to rebase the change I will happily accept a PR from you :beers: cheers!

flosse:
[...]

PR #253, opened 2022-09-07
Gloo net fetch in worker

flosse:
I'm trying to use `gloo-net` in a web extension with manifest V3 but this does not work because of #201.
Fortunately @cdata [found a workaround](https://github.com/cdata/gloo/commit/b23035decc9ce41f40680d52ac7b85add0d00104) :smiley:
That at least avoids a panic but my service worker of the web extension does not have the `DedicadedWorkerGlobalScope` so I changed it to `WorkerGlobalScope` because as far as I know a `DedicadedWorkerGlobalScope` is a specialization of `WorkerGlobalScope`.

What do you think?

### DIFF
--- a/crates/net/src/http/mod.rs
+++ b/crates/net/src/http/mod.rs
@@ -264,18 +264,13 @@ impl Request {
             let window = global.dyn_into::<web_sys::Window>().unwrap();
             window.fetch_with_request(&request)
         } else {
-            let maybe_worker =
-                Reflect::get(&global, &JsValue::from_str("DedicatedWorkerGlobalScope"))
-                    .map_err(js_to_error)?;
+            let maybe_worker = Reflect::get(&global, &JsValue::from_str("WorkerGlobalScope"))
+                .map_err(js_to_error)?;
             if !maybe_worker.is_undefined() {
-                let worker = global
-                    .dyn_into::<web_sys::DedicatedWorkerGlobalScope>()
-                    .unwrap();
+                let worker = global.dyn_into::<web_sys::WorkerGlobalScope>().unwrap();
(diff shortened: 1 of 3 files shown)


crates/net/src/http/mod.rs
```diff
@@ -256,7 +257,25 @@ impl Request {
...
+                .map_err(js_to_error)?;
+            if !maybe_worker.is_undefined() {
+                let worker = global.dyn_into::<web_sys::WorkerGlobalScope>().unwrap();
+                worker.fetch_with_request(&request)
+            } else {
+                return Err(Error::GlooError(
+                    "Unsupported JavaScript global context".into(),
+                ));
+            }
+        };
```
hamza1311, line 277: It's better to `panic!` here and document that. This is a non-recoverable error and returning Result for such error doesn't sound good

hamza1311 (commented):
Can we add test cases for this?

@futursolo do you know if it's possible to run test in gloo-worker?

futursolo:
> Can we add test cases for this?
> 
> @futursolo do you know if it's possible to run test in gloo-worker?

You can send back the result of tests run within workers when code is run normally within a worker.
But you cannot use `#[wasm_bindgen_test]` or `assert*!` in a worker.

crates/net/src/http/mod.rs
flosse: done.

flosse:
@hamza1311 is there something I can do to make this PR merged?

hamza1311:
Sorry, took me a while to get to this. It looks good! Thanks for the PR (and the wait)

PR #255, opened 2022-09-21
gloo_timers, ambiguous verbage

### LINKED ISSUE #254: Cancelable Timeouts/Intervals
## Summary

Synchronously, we get an ID from a `Timeout::new().forget()` or `Interval::new().forget()`.
Is there a reason we have to consume the object and return a u64 ID? I don't think so.

I believe a better solution would be to have an inner field on the object, `id: Option<u64>` which holds such an ID if it was `forgot()`'ed or submitted async. 

And hence, then we could implement `cancel()` after it has been sent as a hook to `web_sys::cancel_timeout_with_handle(&self: &Window, id: u64)`

## Motivation

The motivation is that (in my opinion) it's not straightforward to cancel a Timeout or Interval for the average user, and this is not ergonomic as a result.

## Detailed Explanation

TLDR:

```rust
pub struct Timeout<F: FnOnce> { 
    id: Option<u64>,
    f: F,
}

impl<F: FnOnce> Timeout<F> {
    pub fn new(f: F) -> Self { ... }
    pub fn forget(&mut self) {
        let id = web_sys::submit_timeout.... // Whatever it is currently
        self.id = Some(id)
    }

    pub fn cancel(self) { 
        if let Some(id) = self.id {
            self.id = None;
            web_sys::window().map(|w| w.cancel_timeout_with_handle(id)).expect_throw("No window")
        }
    }

    pub fn id(&self) -> Option<u64> { 
        self.id
    }
}
```

It would be a breaking change.

simbleau:
Closes #254 by improving verbage

### DIFF
--- a/crates/timers/src/callback.rs
+++ b/crates/timers/src/callback.rs
@@ -73,7 +73,7 @@ impl Timeout {
         }
     }
 
-    /// Drops this resource without clearing the timeout.
+    /// Forgets this resource without clearing the timeout.
     ///
     /// Returns the identifier returned by the original `setTimeout` call, and
     /// therefore you can still cancel the timeout by calling `clearTimeout`
--- a/crates/timers/src/callback.rs
+++ b/crates/timers/src/callback.rs
@@ -23,8 +23,8 @@ extern "C" {
 ///
 /// See `Timeout::new` for scheduling new timeouts.
 ///
-/// Resource Acquisition Is Initialization (RAII). Once scheduled, you can
-/// `cancel`/`drop` it or `forget` it to free memory, remaining scheduled.
+/// Once scheduled, you can [`drop`] the [`Timeout`] to clear it or [`forget`](Timeout::forget) to leak it. Once forgotten, the interval will keep running forever.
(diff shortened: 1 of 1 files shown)


crates/timers/src/callback.rs
```diff
@@ -123,8 +125,8 @@ impl Timeout {
 ///
 /// See `Interval::new` for scheduling new intervals.
 ///
-/// Once scheduled, you can either `cancel` so that it ceases to fire or `forget`
-/// it so that it is un-cancel-able.
+/// Resource Acquisition Is Initialization (RAII). Once scheduled, you can
+/// `cancel`/`drop` it or `forget` it to free memory, remaining scheduled.
```
hamza1311, line 129:
```suggestion
/// Once scheduled, you can [`drop`] the [`Interval`] to clear it or [`forget`](Interval::forget) to leak it. Once forgotten, the interval will keep running forever.
/// This pattern is known as Resource Acquisition Is Initialization (RAII).
```

hamza1311 (changes requested):
Thanks for the PR!

simbleau:
Ready for re-review @hamza1311

PR #256, opened 2022-09-30
Fix Request.json(): Use Rust Serde Serialization instead of Javascript…

vpochapuis:
Fix Request.json(): Use Rust Serde Serialization instead of JavaScript Evaluator. Avoids Big Integer serialization issues.
[Related to this issue](https://github.com/yewstack/yew/issues/2894).


hamza1311 (approved):
Thanks for the PR!

Sorry, it took me a while to get to it

PR #264, opened 2022-11-03
fix(ws): calling close event with destroyed close callback

flisky:
I notice the following error:

image

And it's 100% reproduced when `wasm-pack test --chrome`.

It's a racing condition: 

* call `Websocket::close`, which **consumes the ownership**.
* `Websocket::drop` is executed emmediately
  * call `self.ws.close()` the second time (and it doesn't matter).
  * the closures (event listeners) are dropped (and it matters).
* `self.ws.onclose` is called with destroyed `close` callback.

Solution:
* bind`open/close` event listeners once (which I think okay, tell me if not)
* unbind the `open/message/error` eventlisteners when dropping
* trigger `close` event to notify the user (and unbind).

### DIFF
--- a/crates/net/Cargo.toml
+++ b/crates/net/Cargo.toml
@@ -44,11 +44,13 @@ json = ["serde", "serde_json"]
 # Enables the WebSocket API
 websocket = [
     'web-sys/WebSocket',
+    'web-sys/AddEventListenerOptions',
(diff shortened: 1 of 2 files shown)


hamza1311:
Thanks for the PR! Sorry, took me so long to merge this

PR #267, opened 2022-11-05
Function Worker

### LINKED ISSUE #251: Proposal: Function Worker
## Summary

This implemented was originally implemented in https://github.com/yewstack/yew/pull/2773 for Yew Agent.
I think this can also be provided to `gloo-worker` in a framework agnostic fashion (with the exception of SSR).

## Motivation

Previously the `gloo_worker::Worker` type implements an actor model which can be difficult for beginners.

This issue proposes 2 new worker types:

1. `oneshot`: For each input, a single output is produced.
2. `reactor`: A type of worker that operates receives inputs and produces outputs over a channel pair.

## Detailed Explanation

2 Procedural Macros are added:

```rust
#[oneshot]
async fn Squared(input: u32) -> u32 {
    input.pow(2)
}

#[reactor]
async fn SquaredOnDemand(rx: Receiver<u32>, tx: Sender<u32>) {
    while let Some(m) = rx.next().await {
        if tx.send(m.pow(2)).is_err() {
            break;
        }
    }
}
```

2 Spawner types are added as well:

```rust
// OneshotBridge
let squared_bridge = Squared::spawner().spawn();
assert_eq!(squared_bridge.run(2).await, 4);

// ReactorBridge
let (squared_tx, squared_rx) = SquaredOnDemand::spawner().spawn();

squared_tx.send(2);
// Schedules shutdown of bridge.
drop(squared_tx);

assert_eq!(squared_rx.next().await, Some(4));
assert_eq!(squared_rx.next().await, None);
```

### Drawbacks

The procedural macros will use boxed futures until `type_alias_impl_trait` becomes stable.

### Alternative Design

Functional workers can also be implemented using a method like [`tower::service_fn`](https://docs.rs/tower/0.4.13/tower/fn.service_fn.html). With this method, functional workers can be produced without boxing or `type_alias_impl_trait`.

However, this can cause issues with the current spawning method as workers are identified with a struct type.
(The original Agent v2 also uses context providers to identify agents. This method will produce an anonymous type and cannot be used to create context providers.)

## Unresolved Questions

- Should we name the functional workers something else other than `oneshot` and `reactor`?
- Should `gloo_worker::Worker` be renamed to something else to avoid confusion as there will be multiple worker types?

futursolo:
This pull request:

1. Moved actor model workers to `gloo::worker::actor`, exported to `gloo::worker`.
2. Moved most trait requirements from the Worker trait definition to its Spawner and Registrar.
3. Fixed an issue where `Worker::connected` is not being called upon bridge connection.
4. Added `gloo::worker::oneshot` as an 1:1 worker type, constructed via procedural macro.
5. Added `gloo::worker::reactor` and a M:N worker type, constructed via procedural macro.

Still missing:

1. Documentation.
2. Tests for Reactor Worker.

Fix #251.


databasedav:
incredible stuff :pray:

hamza1311:
@futursolo I added docs and test for the reactor worker. Hopefully this PR is ready for merge. If you could review my changes, that would be great

.github/workflows/tests.yml
```diff
@@ -65,6 +65,9 @@ jobs:
             if [[ "$x" == "net" ]]; then
               continue
             fi
+            if [[ "$x" == "worker-macros" ]]; then
+              continue
+            fi
```
futursolo, line 70: There is a CI section for workers below. We should also add the prime example to CI tests.

crates/worker-macros/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+
+[lib]
+proc-macro = true
+
+[dependencies]
+proc-macro-crate = "1.2.1"
+proc-macro-error = "1.0.4"
+proc-macro2 = "1.0.47"
+quote = "1.0.21"
+syn = { version = "1.0.103", features = ["full"] }
```
futursolo, line 21: syn 2 is released. We should upgrade to syn 2.

hamza1311:
@futursolo have a look again

crates/worker-macros/src/worker_fn.rs
```diff
@@ -0,0 +1,280 @@
...
+    pub fn print_inner_fn(&self) -> ItemFn {
+        let mut func = self.func.clone();
+        func.sig.ident = self.inner_fn_ident();
+
+        func.vis = Visibility::Inherited;
+
+        func
+    }
+
+    // Sometimes this users use gloo_worker directly, sometimes it is imported via gloo.
```
futursolo, line 235:
```suggestion
    // Sometimes users use gloo_worker directly, sometimes it is imported via gloo.
```

futursolo (commented):
I cannot approve this PR because I created it.
Other than the mistake in the sentence, which might be due to me, LGTM.

hamza1311:
I'm confused why CI is failing. The test passes for me locally

futursolo:
Maybe try to run CI with Firefox only?

PR #312, opened 2023-02-09
Seanaye/feat/serverside http

seanaye:
This PR adds bindings necessary for server side fetch api. This allows rust code to be deployed easily in v8 isolate environments like cloudflare workers or Deno Deploy

### DIFF
--- a/crates/net/src/http/mod.rs
+++ b/crates/net/src/http/mod.rs
@@ -20,6 +20,7 @@ use crate::{js_to_error, Error};
 use js_sys::Reflect;
 use js_sys::{ArrayBuffer, Uint8Array};
 use std::fmt;
+use std::ops::DerefMut;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
 use wasm_bindgen_futures::JsFuture;
@@ -72,14 +73,104 @@ impl fmt::Display for Method {
 }
 
 /// A wrapper round `web_sys::Request`: an http request to be used with the `fetch` API.
-pub struct Request {
+pub struct RequestWritable {
     options: web_sys::RequestInit,
     headers: Headers,
     query: QueryParams,
     url: String,
 }
 
-impl Request {
+/// A writable wrapper around `web_sys::Reponse`: an http response to be used with the `fetch` API
+/// on a server side javascript runtime
+#[derive(Debug)]
+pub struct ResponseWritable {
+    headers: Headers,
+    options: web_sys::ResponseInit,
+    body: Option<ResponseBody>,
+}
+
+/// Possible initializers for request body
+#[derive(Debug)]
+pub enum ResponseBody {
+    /// Blob response body
+    Blob(web_sys::Blob),
+    /// Buffer response body
+    Buffer(js_sys::Object),
+    /// `Uint8Array` response body
+    U8(Box<[u8]>),
+    /// `FormData` response body
+    Form(FormData),
+    /// `URLSearchParams` response body
+    Search(web_sys::UrlSearchParams),
+    /// String response body
+    Str(String),
+    /// ReadableStream response body
+    Stream(web_sys::ReadableStream),
+}
+
+impl ResponseWritable {
+    /// Creates a new response object
+    pub fn new() -> Self {
+        Self {
+            headers: Headers::new(),
+            options: web_sys::ResponseInit::new(),
+            body: None,
+        }
+    }
+
+    /// Replace _all_ the headers.
+    pub fn headers(mut self, headers: Headers) -> Self {
+        self.headers = headers;
+        self
+    }
+
+    /// Sets a header.
+    pub fn header(self, key: &str, value: &str) -> Self {
+        self.headers.set(key, value);
+        self
+    }
+
+    /// Set the status code
+    pub fn status(mut self, status: u16) -> Self {
+        self.options.status(status);
+        self
+    }
+
+    /// Set the status text
+    pub fn status_text(mut self, status_text: &str) -> Self {
+        self.options.status_text(status_text);
+        self
+    }
+
+    /// Set the response body
+    pub fn body(mut self, data: ResponseBody) -> Self {
+        self.body = Some(data);
+        self
+    }
+
+    /// Consume the wrapper and return a result of http response object
+    pub fn into_raw(mut self) -> Result<web_sys::Response, Error> {
+        use web_sys::Response as R;
+        self.options.headers(&self.headers.into_raw());
+        let init = &self.options;
+        match self.body {
+            None => R::new_with_opt_str_and_init(None, init),
+            Some(x) => match x {
+                ResponseBody::Blob(y) => R::new_with_opt_blob_and_init(Some(&y), init),
+                ResponseBody::Buffer(y) => R::new_with_opt_buffer_source_and_init(Some(&y), init),
+                ResponseBody::U8(mut y) => R::new_with_opt_u8_array_and_init(Some(y.deref_mut()), init),
+                ResponseBody::Form(y) => R::new_with_opt_form_data_and_init(Some(&y), init),
+                ResponseBody::Search(y) => {
+                    R::new_with_opt_url_search_params_and_init(Some(&y), init)
+                }
+                ResponseBody::Str(y) => R::new_with_opt_str_and_init(Some(&y), init),
+                ResponseBody::Stream(y) => R::new_with_opt_readable_stream_and_init(Some(&y), init),
+            },
+        }.map_err(js_to_error)
+    }
+}
+
+impl RequestWritable {
     /// Creates a new request that will be sent to `url`.
     ///
     /// Uses `GET` by default. `url` can be a `String`, a `&str`, or a `Cow<'a, str>`.
@@ -247,7 +338,7 @@ impl Request {
     }
 
     /// Executes the request.
-    pub async fn send(mut self) -> Result<Response, Error> {
+    pub async fn send(mut self) -> Result<ResponseReadable, Error> {
         let url = String::from(&self.build_url()?.to_string());
         self.options.headers(&self.headers.into_raw());
         let request =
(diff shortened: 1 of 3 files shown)


hamza1311:
I don't understand the use case. Why do you need to manually construct `Response`?

seanaye:
This is required for use in modern serverside js runtimes, currently there is no way to build a response to the client in gloo.

See https://github.com/seanaye/discord-bot/blob/main/main.ts

To run that example locally `deno task dev`

seanaye:
I was just looking at the docs for [http](https://docs.rs/http/latest/http/index.html) it looks like it would be possible to implement `impl TryInto<web_sys::Response> for http::Response` as an alternative

seanaye:
I have a new proposed api which leverages try_into and try_from for Request and Response gonna mark as draft until I finish the docs and tests but the code is mostly done

crates/net/src/http/response.rs
```diff
@@ -0,0 +1,266 @@
...
+
+/// The [`Request`]'s response
+pub struct Response {
+    raw: web_sys::Response,
+}
+
+impl Response {
+    /// Returns an instance of response builder
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new() -> Builder {
```
futursolo, line 26:
```suggestion
    pub fn builder() -> Builder {
```

futursolo:
In general, I think this is a good change which gives requests a separate builder for builder pattern and requests becomes a more general reader / writer type.

However, I am a little bit unsure how this could actually be used by actual applications as I do not think users can avoid runtime specific binding for more complex function as each runtime is different on how to access additional information that describes the function itself (commonly referred to as ["Context"](https://docs.rs/lambda_runtime/latest/lambda_runtime/struct.Context.html)) and how to access environment variables or secrets.

E.g.: Cloudflare Workers provide [`cf()`](https://docs.rs/worker/latest/worker/struct.Request.html#method.cf) method to access additional request information not provided by `web_sys::Request` and [`worker::Env`](https://docs.rs/worker/latest/worker/struct.Env.html) to access environment variables and secrets.

seanaye:
@futursolo the env variables can be passed in as second arguments to the handler function if required. I actually have a live demo of this code working on Deno Deploy if you want to see. I'm not passing any env vars but it would also be possible.

Live demo link: https://seanaye-discord-bot.deno.dev/
Source for demo: https://github.com/seanaye/discord-bot/blob/main/main.ts

futursolo:
> Lambda's are not a fetch web standard based API. Deno Deploy doesnt really have a Context it does have environment variables but these can be passed as function arguments. 

According to the documentation, environment variables are accessed via [`Deno.env`](https://deno.com/deploy/docs/runtime-api#deno-apis) which is a runtime-specific API.
Information that describes the function itself is [provided](https://deno.com/deploy/docs/projects#preset-variables) as environment variables as well.

https://deno.com/deploy/docs/projects#environment-variables

> It looks like cloudflare workers dont abide by the Request standard, but Deno does in Deno Deploy

I do not think there is a "standard" of how `Request` and `Response` should be used if they are not paired with fetch. They are defined as input and results of the `fetch()` API [Specification](https://fetch.spec.whatwg.org/). Additional information of the request in Deno Deploy is provided as the second argument to the [function](https://deno.land/std@0.140.0/http/server.ts?s=Handler) as [`ConnInfo`](https://deno.land/std@0.140.0/http/server.ts?s=ConnInfo), which is runtime specific.

seanaye:
ah I see what you mean. I still think this is much better for handling Request/Response on the server with wasm. I don't think there is a good solution for handling things like `Deno.env` outside of passing it to the handler as a second argument e.g.

```ts
serve((req, connInfo) => handler(req, Deno.env.get("MY_API_SECRET"), connInfo.remoteAddr.definition.transport)
```

```rust
#[wasm_bindgen]
pub async fn handler(req: web_sys::Resquest, secret: String, transportMethod: String) -> Result<web_sys::Response, JsError> {
...
}

These values are platform specific as you mentioned and should not be handled by gloo. The caller should be responsible for them. Its possible to injects the values into rust as other arguments

That being said 99% of applications dont need to use things like `connInfo`, most relevant info is available inside the `Request` standard object as defined in the spec

**Comment by seanaye** [comment]
Is there anything further you need from me to move this forward?

File: crates/net/src/http/response.rs [hunk 2]
```diff
@@ -0,0 +1,268 @@
...
+            body: None,
+        }
+    }
+}
+
+impl fmt::Debug for ResponseBuilder {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Response Builder")
+            .field("headers", &self.headers)
+            .finish()
```
futursolo, line 266:
```suggestion
            .finish_non_exhaustive()
```

crates/net/src/http/response.rs
```diff
@@ -0,0 +1,268 @@
...
+    #[cfg(feature = "json")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
+    pub fn json<T: serde::Serialize + ?Sized>(self, value: &T) -> Result<Response, Error> {
+        let json = serde_json::to_string(value)?;
+        self.header("Content-Type", "application/json")
+            .body(Some(ResponseBody::Str(json)))
+    }
+
+    /// Set the response body and return the response
+    pub fn body(mut self, data: Option<ResponseBody>) -> Result<Response, Error> {
```
futursolo, line 225:
```suggestion
    pub fn body<T>(mut self, body: T) -> Result<Response, Error>
    where
        T: IntoRawResponse
    {

pub trait IntoRawResponse {
    fn into_raw_response(init: ResponseInit) -> web_sys::Response;
}
```

Instead of using an enum, I would suggest we allow any type that implements a certain trait (I have named it IntoRawResponse in this example) to be used as body.
This way we can avoid the full match below, which should also provide better dead code elimination.

crates/net/src/http/response.rs
```diff
@@ -0,0 +1,268 @@
...
+            .finish()
+    }
+}
+
+/// A writable wrapper around `web_sys::Reponse`: an http response to be used with the `fetch` API
+/// on a server side javascript runtime
+pub struct ResponseBuilder {
+    headers: Headers,
+    options: web_sys::ResponseInit,
+    body: Option<ResponseBody>,
```
futursolo, line 159: I don't think the body field needs to be saved. We can get a smaller struct size by removing the body method.

futursolo (commented):
I am sorry that I was in the process of switching to a new PC. So I wasn't able to review this pull request.

I have put in some suggestions. Hope we can get this merged soon.

seanaye:
Thanks for the review, somehow didn't see it until now. Will probably get a chance to implement later this week

seanaye:
@futursolo should be good to re-review / merge now

crates/net/src/http/response.rs
```diff
@@ -0,0 +1,281 @@
...
+impl IntoRawResponse for Option<&web_sys::ReadableStream> {
+    fn into_raw(self, init: ResponseInit) -> Result<web_sys::Response, JsValue> {
+        web_sys::Response::new_with_opt_readable_stream_and_init(self, &init)
+    }
+}
+
+/// trait which allow consuming self into a raw web_sys::Response
+pub trait IntoRawResponse {
+    /// function which converts self and a web_sys::ResponseInit into a result to a
+    /// web_sys::Response
```
futursolo, line 262:
```suggestion
    /// A method which converts `self` and a [`web_sys::ResponseInit`] into a result to a
    /// [`web_sys::Response`].
```

crates/net/src/http/response.rs
```diff
@@ -0,0 +1,281 @@
...
+        Self {
+            headers: Headers::new(),
+            options: web_sys::ResponseInit::new(),
+        }
+    }
+}
+
+impl fmt::Debug for ResponseBuilder {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Response Builder")
```
futursolo, line 277:
```suggestion
        f.debug_struct("ResponseBuilder")
```

The name provided to the debug implementation should match the name of the struct.

crates/net/src/http/request.rs
```diff
@@ -0,0 +1,368 @@
...
+        self
+    }
+
+    /// Sets the request abort signal.
+    pub fn abort_signal(mut self, signal: Option<&AbortSignal>) -> Self {
+        self.options.signal(signal);
+        self
+    }
+
+    pub async fn send(self) -> Result<Response, Error> {
```
futursolo, line 181:
```suggestion
    /// Builds the request and send it to the server, returning the received response.
    pub async fn send(self) -> Result<Response, Error> {
```

crates/net/src/http/request.rs
```diff
@@ -0,0 +1,368 @@
...
+        let req: Request = self.try_into()?;
+        req.send().await
+    }
+
+    pub fn build(self) -> Result<Request, crate::error::Error> {
```
futursolo, line 186:
```suggestion
    /// Builds the request.
    pub fn build(self) -> Result<Request, crate::error::Error> {
```

crates/net/src/http/mod.rs
```diff
@@ -15,434 +15,10 @@
...
-        f.debug_struct("Response")
-            .field("url", &self.url())
-            .field("redirected", &self.redirected())
-            .field("status", &self.status())
-            .field("headers", &self.headers())
-            .field("body_used", &self.body_used())
-            .finish()
-    }
-}
+pub use request::Request;
```
futursolo, line 23:
```suggestion
#[doc(inline)]
pub use request::Request;
```

We should copy the documentation from the http crate to gloo-net.

futursolo (commented):
Thank you.

I think the implementation looks good.

I have added some minor suggestions for documentation and I think it would be mergable after they are fixed.

I will also request @hamza1311 to have a look at this PR as they are the original author.

futursolo (approved):

PR #332, opened 2023-05-01
`net` improve interoperability of wrapper types

Alextopher:
# `net` improve interoperability of wrapper types

Specifically [C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits).

This PR currently only covers the `gloo_net` crate. I'd be happy to (slowly) extend this idea to other crates if this is well received.

### `Request`

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_circle: | The original reason I started this PR. It can be useful to clone a `Request` to send it multiple times. |
| Copy | :x: | The type is large (40 bytes) and copying should be explicit, and `web_sys` types don't implement `Copy` |
| PartialEq | :green_circle: | It would be odd to check if you have two identical requests, but there's no reason to forbid it. |
| Eq | :green_circle: | See `PartialEq`, reflexivity is guaranteed. |
| PartialOrd | :x: | Not applicable. |
| Ord | :x: | Not applicable. |
| Hash | :x: | Inner web_sys types don't implement `Hash` |
| Debug | :green_square: | Existing.
| Display | :x: | Requests are fairly complex, perhaps a `Display` implementation could be implemented to follow the HTTP spec, but that's a larger discussion. |
| Default | :x: | No obvious default value. |

### `Response`

Implement the same traits as the underlying `web_sys` type.

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_circle: | Added. |
| Copy | :x: | Inner `web_sys` types don't implement `Copy` |
| PartialEq | :green_circle: | Added. |
| Eq | :green_circle: | Added. |
| PartialOrd | :x: | Not applicable. |
| Ord | :x: | Not applicable. |
| Hash | :x: | Inner web_sys types don't implement `Hash` |
| Debug | :green_square: | Existing.
| Display | :x: | Perhaps we could follow the HTTP spec? |
| Default | :x: | No obvious default value. |

### Headers

Implement the same traits as the underlying `web_sys` type.

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_circle: | Added. |
| Copy | :x: | |
| PartialEq | :green_circle: | Added. |
| Eq | :green_circle: | Added. |
| PartialOrd | :x: | |
| Ord | :x: | |
| Hash | :x: | |
| Debug | :green_square: | Existing. |
| Display | :x: | Could be implemented to follow the HTTP spec. |
| Default | :green_square: | Existing. |

### `QueryParams`

Implement the same traits as the underlying `web_sys` type.

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_circle: | Added. |
| Copy | :x: | |
| PartialEq | :green_circle: | Added. |
| Eq | :green_circle: | Added. |
| PartialOrd | :x: | |
| Ord | :x: | |
| Hash | :x: | |
| Debug | :green_square: | Existing. |
| Display | :green_square: | Existing. |
| Default | :green_square: | Existing. |

#### `Method`

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_square: | Existing. |
| Copy | :green_square: | Existing. |
| PartialEq | :green_circle: | A simple Enum over the types of HTTP methods should be comparable. |
| Eq | :green_circle: | Method::X == Method::X |
| PartialOrd | :green_circle: | I've implemented the lexicographical ordering. I added a test against the `Display` implementation |
| Ord | :green_circle: | See `PartialOrd` |
| Hash | :green_circle: | A simple Enum is hashable. |
| Debug | :green_square: | Existing. |
| Display | :green_square: | Existing. |
| Default | :x: | Not applicable. |

Added `as_str` method to `Method` to get a `&'static str` representation of the method. This avoids some unnecessary allocations.

### `EventSource`

Implement the same traits as the underlying `web_sys` type.

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_square: | Existing. |
| Copy | :x: | |
| PartialEq | :green_circle: | Added. |
| Eq | :green_circle: | Added. |
| PartialOrd | :x: | |
| Ord | :x: | |
| Hash | :x: | |
| Debug | :green_square: | Existing. |
| Display | :x: | |

### `EventSourceSubscription`

I don't think any traits are applicable here. `Closure` prevents most traits from being implemented.

| trait |  | reason |
|:-----|:----|:----|
| Debug | :green_square: | Existing. |

### `eventsource::State`

In JS the `readyState` property returns a Number 0, 1, or 2 (connecting, open, closed). I've implemented the expected traits for this type.

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_square: | Existing. |
| Copy | :green_square: | Existing. |
| PartialEq | :green_circle: | Added. |
| Eq | :green_circle: | Added. |
| PartialOrd | :green_circle: | Added. |
| Ord | :green_circle: | Added. |
| Hash | :green_circle: | Added. |
| Debug | :green_square: | Existing. |
| Display | :green_circle: | "connecting", "open", "closed" |

Added `as_str` to get a `&'static str` representation of the state. This avoids some unnecessary allocations.

Added `From<u16> for State` and `From<State> for u16` to convert between the JS number and the Rust enum.

### `EventSourceError`

I've left this trait unchanged, since its `non_exhaustive` I'm not confident if it will remain as simple as it is now. If it does remain "simple" I'd suggest implementing `Copy` and `Hash`.

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_square: | Existing. |
| Copy | :warning: | If this type will remain simple then it should become copyable. |
| PartialEq | :green_circle: | Added. |
| Eq | :green_circle: | Added. |
| PartialOrd | :x: | Error types aren't orderable |
| Ord | :x: | |
| Hash | :warning: | If this type will remain simple then it should become hashable. |
| Debug | :green_square: | Existing. |
| Display | :green_square: | Existing. |

`Ord` doesn't make sense for an error type.

### `CloseEvent`

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_square: | Existing. |
| Copy | :x: | 32 bytes is a bit large for `Copy` |
| PartialEq | :green_circle: | Added. |
| Eq | :green_circle: | Added. |
| PartialOrd | :x: | |
| Ord | :x: | |
| Hash | :green_circle: | Added. |
| Debug | :green_square: | Existing. |
| Display | :x: | I think individual users will have differing preferences for how to display this type. |

### `WebSocket`

No traits are applicable here. The underlying types prevent most traits from being implemented. Particularly the `Closure` fields.

### `Message`

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_square: | Existing. |
| Copy | :x: | `String` and `Vec` aren't `Copy` |
| PartialEq | :green_circle: | Existing. |
| Eq | :green_square: | Existing. |
| PartialOrd | :x: | Not applicable. |
| Ord | :x: | Not applicable. |
| Hash | :green_circle: | Added. |
| Debug | :green_square: | Existing. |
| Display | :x: | I think individual users will have differing preferences for how to `Display` a `Vec<u8>` |

### `websocket::State`

Similar to eventsource::State.

In JS the `readyState` property returns a Number (0, 1, 2, 3) (connecting, open, closing, closed). I've implemented the expected traits for this type.

| trait |  | reason |
|:-----|:----|:----|
| Clone | :green_square: | Existing. |
| Copy | :green_square: | Existing. |
| PartialEq | :green_circle: | Added. |
| Eq | :green_circle: | Added. |
| PartialOrd | :green_circle: | Added, follows the ordering of connection lifecycle |
| Ord | :green_circle: | Added, follows the ordering of connection lifecycle |
| Hash | :green_circle: | Added. |
| Debug | :green_square: | Existing. |
| Display | :green_circle: | "connecting", "open", "closing", "closed" |

Added `as_str` to get a `&'static str` representation of the state. This avoids some unnecessary allocations.

Added `From<u16> for State` and `From<State> for u16` to convert between the JS number and the Rust enum.

### `WebSocketError`

This enum depends on `JsError` which does not implement many traits. If that changes then this type should implement more traits.

| trait |  | reason |
|:-----|:----|:----|
| Clone | :x: | `JsError` doesn't implement `Clone` |
| Copy | :x: | `JsError` doesn't implement `Copy` |
| PartialEq | :x: | `JsError` doesn't implement `PartialEq` |
| Eq | :x: | `JsError` doesn't implement `Eq` |
| PartialOrd | :x: | Error types aren't orderable |
| Ord | :x: | |
| Hash | :x: | `JsError` doesn't implement `Hash` |
| Debug | :green_square: | Existing. |
| Display | :green_square: | Existing. |

### `Error`

Same as `WebSocketError`, `JsError` hasn't been covered by this PR. Once `JsError` implements more traits this type should implement more traits.

| trait |  | reason |
|:-----|:----|:----|
| Debug | :green_square: | Existing. |


futursolo (changes requested):
Thank you for the pull request. I have only reviewed the first section (http) as the master branch has conflicts with this pull request.

I do wish to provide some of my personal opinion on the first part of the implementation which I think is applicable for the rest of the pull request as well.

Feel free to request another review after you have resolved the conflicts.

### Request

| <div style="width:290px">Trait</div> | My Opinion | Reason |
-- | -- | --
Clone | ❌ | I do not think `web_sys::RequestInit` is clonable when `web_sys::ReadableStream` is used as body. We shouldn't provide an implementation that could panic from time to time.
PartialEq | ❌ | web_sys delegates them to `a == b` in JavaScript and not trying to check whether the underlying value is equal. I do not think this provides the desired behaviour when `==` is used in Rust where users would expect their internal values to equal. image <br> I do not think a sound `PartialEq` and `Eq` implementation is possible with `web_sys::RequestInit` as you cannot compare `web_sys::ReadableStream` if it is used as body for one or both requests used in the comparison. 
Eq | ❌ | See above

### Response

Similar to request.

### Headers

Trait | My Opinion | Reason
-- | -- | --
Clone | ⚠️ | I think this can be provided. However, we should not use the web_sys's Clone implementation. web_sys only increases the reference count of the same `Headers` JavaScript instance and not actually cloning the value into a different `Header()`. This would cause the cloned headers instance share any changes applied on the original header instance after the clone. If you wish to implement Clone, you need to implement them manually with `Headers::new_with_headers()`.
PartialEq | ⚠️ | This is similar to request where `a == b` does not provide desirable result, but I think it's not as difficult to provide a manual implementation. If you wish to implement PartialEq, you need to implement the trait manually. image
Eq | ⚠️ | See above

### QueryParams

Similar to `Headers`.

### Method

We switched to `Method` type from the `http` crate. This is now only a type alias.


#### Legend

❌: I do not think a sound implementation is possible.
⚠️: I think a sound implementation is possible but we should do something else than what web_sys is doing.

Alextopher:
Just so that I understand, the `web_sys` `#[derive(Clone)]` macro is not the same as calling the object's `clone` method in JS (for example https://fetch.spec.whatwg.org/#dom-request-clone)? 

So to improve this we either
- Need to write our own way of calling `clone` (using wasm_bindgen) and then write a safe wrapper (something like `TryClone`) to catch the TypeError
- Write a competing clone method that "builds" identical objects through the existing `web_sys` bindings

futursolo:
> Just so that I understand, the web_sys #[derive(Clone)] macro is not the same as calling the object's clone method in JS (for example https://fetch.spec.whatwg.org/#dom-request-clone)?

In this case, the `Request` struct holds a `web_sys::RequestInit` instead of `web_sys::Request`. `web_sys::RequestInit` is the `RequestInit` dictionary described in the spec, which should be a plain JavaScript object that doesn't have a `clone` method.

The derived Clone method for `RequestInit` eventually gets delegated to web_sys::JsValue which then simply increases the object reference count in a special array used by wasm_bindgen to access JavaScript values.

See: https://rustwasm.github.io/wasm-bindgen/api/src/wasm_bindgen/lib.rs.html#1091-1099


> Need to write our own way of calling clone (using wasm_bindgen) and then write a safe wrapper (something like TryClone) to catch the TypeError
> Write a competing clone method that "builds" identical objects through the existing web_sys bindings

I am fine with both ways as long as it does not significantly increase the bundle size.
(I am also fine with doing away with the clone implementation completely if it ends up too complicated.)

However, I do not think a non-fallible `clone` is possible as `web_sys::ReadableStream` is not clonable and any requests that use it as body will in turn become unclonable.

In the latest master branch, the Request struct is separated into Request and RequestBuilder, it might be easier if only implementing cloning on RequestBuilder as the builder does not contain a body?

crates/net/src/http/headers.rs
```diff
@@ -6,6 +6,7 @@ use wasm_bindgen::{JsCast, UnwrapThrowExt};
 // I experimented with using `js_sys::Object` for the headers, since this object is marked
 // experimental in MDN. However it's in the fetch spec, and it's necessary for appending headers.
 /// A wrapper around `web_sys::Headers`.
+#[derive(Clone, PartialEq, Eq)]
```
hamza1311, line 9:
```rust
let headers = todo!()
let headers2 = headers.clone()
headers2.append("key123", "value")

assert!(!headers.has("key123"))
```
Can you check that this above snippet doesn't panic? I want to ensure that we're not sharing the same instance of `web_sys::Headers`

hamza1311 (commented):
Thanks for the PR!

This was mostly oversight when implementing the types. Thanks for addressing it

crates/net/src/http/headers.rs
Alextopher: I agree with @futursolo that derives here are most likely unsound. Now I understand the `web_sys` `Clone, PartialEq, Eq` implementations call out to JS to preform the shallow copies and comparisons

crates/net/src/http/headers.rs
```diff
@@ -37,19 +29,33 @@ impl Headers {
...
+    /// # fn no_run() {
+    /// let headers = Headers::new();
+    /// headers.append("Content-Type", "text/plain");
+    /// assert_eq!(headers.get("Content-Type"), Some("text/plain".to_string()));
+    ///
+    /// headers.append("Content-Type", "text/html");
+    /// assert_eq!(headers.get("Content-Type"), Some("text/plain, text/html".to_string()));
+    /// # }
+    /// ```
+    pub fn append(&mut self, name: &str, value: &str) {
```
Alextopher, line 46: I know `web_sys::Headers` (and all? `web_sys` types)  allow inner mutability through a `&T` but I'm not sure if it's idiomatic to do the same on our wrapper types. 

For example, if someday we choose to change the implementation of `Header` wrap a `HashSet<&str, String>`  this method would need to be `&mut` or we're committing to using a `RefCell<_>` forever

crates/net/src/http/request.rs
```diff
@@ -38,10 +38,9 @@ impl RequestBuilder {
     }
 
     /// Set the body for this request.
-    pub fn body(mut self, body: impl Into<JsValue>) -> Result<Request, Error> {
+    pub fn body(mut self, body: impl Into<JsValue>) -> Self {
```
Alextopher, line 41: I didn't see a reason why Request `body` should `build` , this allows us to set the body in any order.

I will add a test to verify this is sound.

crates/net/src/http/response.rs
```diff
@@ -78,43 +93,93 @@ impl Response {
...
+    ///
+    /// # Note
+    ///
+    /// This consumes the response, if you need to access the body multiple times,
+    /// you should `clone` the response first.
+    ///
+    /// # Errors
+    ///
+    /// This method may return a "RangeError"
+    pub async fn binary(self) -> Result<Vec<u8>, Error> {
```
Alextopher, line 182: Methods that consume Request/Response bodies now take ownership of `self`. This makes `clone` infallible and prevents the class of bugs where users read the body multiple times.

Alextopher:
I haven't gone back to websocket or eventsource yet.

futursolo:
However, after re-examining the updated pull request, some of criteria has exceeded simply adding additional traits and I think is kind of difficult to review them with all the changes mixed in the same PR.

Could you please split the pull request into multiple smaller ones and limit to adding traits?

I think the API changes contained in this pull request to certain types are significant, if you wish to change them, would you mind start with an [RFC](https://github.com/rustwasm/gloo/issues/new?assignees=&labels=&template=propose_design.md&title=)?

Alextopher:
(draft) RFC: #335 I agree the changes I've suggested have surpassed what this PR was originally for. I think it's worth closing :smiley:

PR #333, opened 2023-05-09
Modularizes gloo

xfbs:
This turns the individual subcrates into optional features. By default, all features are enabled, to ensure this change is not a breaking change. But if needed, default features can be turned off and only the modules needed enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,25 +12,51 @@ documentation = "https://docs.rs/gloo/"
 categories = ["api-bindings", "wasm"]
 
 [dependencies]
-gloo-timers = { version = "0.2", path = "crates/timers" }
-gloo-events = { version = "0.1", path = "crates/events" }
-gloo-file = { version = "0.2", path = "crates/file" }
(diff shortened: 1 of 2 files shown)


hamza1311:
Thanks for the PR! I have to ask: why make this change? Depending upon the individual crates is always an option

xfbs:
Hey!

I can try to explain:

Basically, right now with `gloo` you either get *everything* or *nothing* (you have to manually import crates). Often times it would be convenient to import `gloo` and have something in-between. For example get gloo with the `net`, `timers` and `events` features. Somewhat like `tokio` (where you would enable `fs` and `sync` features).

Especially in WASM where you are trying to get small binaries (because they load through the browser) you often don't want to blindly import everything (although, to be fair, `wasm-opt` seems to be doing a really good job of throwing stuff that isn't used out).

If you think this does not make sense, feel free to disregard it. I would like for `gloo` to be modular like this, but it is not really "necessary", just a way for lazy people like me to use `gloo` without needing to add many manual imports! 

Cheers,
-Patrick

hamza1311 (approved):
Sounds reasonable

PR #334, opened 2023-05-10
Seanaye/feat/add misc methods

seanaye:
Adds a number of utility methods that I have needed while working on a server integration

* Getting properties from the `ResponseBuilder`
* Access to the Response.clone javascript method [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Response/clone)
* Clone trait for `ResponseBuilder`

### DIFF
--- a/crates/net/src/http/request.rs
+++ b/crates/net/src/http/request.rs
@@ -343,6 +343,11 @@ impl Request {
             .map_err(|e| panic!("fetch returned {:?}, not `Response` - this is a bug", e))
             .map(Response::from)
     }
+    /// attempts to clone the request via the Request.clone api
+    pub fn try_clone(&self) -> Result<Self, Error> {
+        let clone = self.0.clone().map_err(js_to_error)?;
+        Ok(Self(clone))
+    }
 }
 
 impl From<web_sys::Request> for Request {
--- a/crates/net/src/http/response.rs
+++ b/crates/net/src/http/response.rs
@@ -177,6 +177,23 @@ impl ResponseBuilder {
         self
     }
 
+    /// Get a reference to the contained headers
+    pub fn get_headers(&self) -> &Headers {
+        &self.headers
+    }
+
+    /// Get the contained status code if it exists
+    pub fn get_status(&self) -> Option<f64> {
+        js_sys::Reflect::get(&self.options, &JsValue::from_str("status")).ok()?.as_f64()
+    }
+
+    /// Get the contained status text if it exists
+    pub fn get_status_text(&self) -> Option<String> {
+        js_sys::Reflect::get(&self.options, &JsValue::from_str("statusText"))
+            .ok()?
+            .as_string()
+    }
+
     /// Set the status code
     pub fn status(mut self, status: u16) -> Self {
         self.options.status(status);
@@ -278,4 +295,16 @@ impl fmt::Debug for ResponseBuilder {
             .field("headers", &self.headers)
             .finish_non_exhaustive()
     }
+}
+
+impl Clone for ResponseBuilder {
+    fn clone(&self) -> Self {
+        let headers = Headers::new();
+        for (name, value) in self.headers.entries() {
+            headers.append(name.as_str(), value.as_str())
+        }
+        let options = self.options.clone();
+        Self { headers, options }
+    }
+
 }
--- a/crates/net/src/http/mod.rs
+++ b/crates/net/src/http/mod.rs
@@ -24,4 +24,4 @@ pub use http::Method;
 pub use query::QueryParams;
 
 pub use request::Request;
-pub use response::{IntoRawResponse, Response};
+pub use response::{IntoRawResponse, Response, ResponseBuilder};


Alextopher:
I've been dragging my feet on writing an RFC on these types. I'm a bit concerned with how `Request` doesn't follow typical ownership and borrowing rules (you only need a &T even when mutating) and I've noticed `RequestBuilder` is missing a lot of options. Here is how I've been thinking of improving this:

In order for a method like `Request::url(&self) -> &String` to exist we need something in rust to own the url in Rust.  The same thing goes with the other getters. Instead of wrapping `web_sys` types, I think we should keep values in Rust as long as possible.

```rust
/// RequestOptions is used to create a [`web_sys::RequestInit`]. 
struct RequestOptions {
    method: http::Method,
    headers: http::HeaderMap, // or our current `Headers` type
    body: Option<Body>,
    cache: web_sys::RequestCache, // These web_sys types are enums
    credentials: web_sys::RequestCredentials,
    integrity: String,
    mode: web_sys::RequestMode,
    redirect: web_sys::RequestRedirect,
    referrer: String,
    referrer_policy: web_sys::ReferrerPolicy,
    // signal: Option<&'a web_sys::AbortSignal>,
}

/// A convenient builder for [`Request`].
#[derive(Debug)]
#[must_use = "RequestBuilder does nothing unless you assign a body or call `build`"]
pub struct RequestBuilder {
    url: String,
    init: RequestOptions,
}

/// A wrapper around [`web_sys::Request`]. Requests are immutable after they are created.
#[derive(Debug)]
pub struct Request {
    inner: web_sys::Request, // Only used to `send` and process the `body`
    // Cached Rust representations of the request inners.
    url: String,
    init: RequestOptions,
}
```

I think this creates cleaner code: https://github.com/Alextopher/gloo/blob/http/crates/net/src/http/request.rs

Alextopher:
The problem I see with my suggestion: it's a lot of breaking changes. However, I think it makes the library more mature and idiomatic and opens up more features that you previously had to use `web_sys` to use.

Alextopher:
#335 my current draft

seanaye:
Could you please provide an example of where a mutation occurs with &T for the current code? I can't find anything other than reading the body which is a bit of a special case imo. I agree the builder needs the missing config fields. I looked at your  PR I don't have any concerns other than breaking changes

Alextopher:
I think methods that consume the body should capture ownership over self, reading the body twice (without cloning) is almost always bug.

But you're right my mutability concerns are mostly in QueryParams and Headers https://github.com/rustwasm/gloo/blob/master/crates/net/src/http/headers.rs#L61

seanaye:
I don't think its required to keep the init options inside of the `Request`. Its possible to keep it as `struct Request(web_sys::Request)` and use `js_sys::Reflect::get` to access the init options e.g. to access `mode`. Once the `Request` is built all properties are read only

seanaye:
Re: the duplicated reading, I think this is a matter of preference. I agree if the body was its own struct then reading it should take ownership but there is a lot of other data associated with the `Request`. It doesn't make sense to me that you cant read the url after reading the body because reading the body consumed the entire request.

Alextopher:
I see that, I think there's a tradeoff:

Do we want this signature (which makes the returned value immutable)
```rust
    pub fn integrity(&self) -> &String {
        &self.options.integrity
    }
```
Or is this good enough?
```
    pub fn integrity(&self) -> String {
        self.raw.integrity()
    }
```

There's also a small time/space tradeoff. Using some extra space in wasm world can asymptomatically reduce the number of calls from wasm to js. What I've been able to do today : but I haven't pushed yet : is remove the the `web_sys::Request` from `Request` and only build it when needed (when calling `send()` or parsing the body).

```rust
#[derive(Debug)]
pub struct Request {
    url: String,
    options: RequestOptions,
}

impl From<web_sys::Request> for Request {
    fn from(value: web_sys::Request) -> Self {
        Self {
            url: value.url(),
            options: RequestOptions {
                method: http::Method::from_bytes(value.method().as_bytes()).unwrap(),
                headers: headers_from_js(&value.headers()),
                body: value.body().map(Body::from),
                cache: value.cache(),
                credentials: value.credentials(),
                integrity: value.integrity(),
                mode: value.mode(),
                redirect: value.redirect(),
                referrer: value.referrer(),
                referrer_policy: value.referrer_policy(),
            },
        }
    }
}

impl TryFrom<Request> for web_sys::Request {
    type Error = crate::Error;

    fn try_from(value: Request) -> Result<Self, Self::Error> {
        let init: web_sys::RequestInit = value.options.into();
        web_sys::Request::new_with_str_and_init(value.url.as_str(), &init).map_err(js_to_error)
    }
}
```

Alextopher:
> Re: the duplicated reading, I think this is a matter of preference. I agree if the body was its own struct then reading it should take ownership but there is a lot of other data associated with the Request. 

I dislike how it's possible to read the body multiple times. In `reqwest` they use the same pattern https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.text for the same reasoning

I can think of changing the signature of send to return a tuple:

```rust
struct Body(web_sys::ReadableStream);

impl Body {
    async fn text(self) -> Result<String>; 
    // etc
}

impl Request {
    fn send(self) -> Result<(Response, Body)> 
}
```

seanaye:
I think taking ownership is fine actually, just forces users to get what they want out of the response before reading the body. I will add a commit to take the ownership

seanaye:
Do you think `fn body_used` should be removed now? its not possible to read a body multiple times

Alextopher:
I'm not sure - there's probably some value of it with regards to backwards compatibility, Or if someone is making a lot of use of `into_raw()` and `from_raw()`. I'd add a note that in normal operation it will return false

crates/net/src/http/response.rs
```diff
@@ -125,6 +125,12 @@ impl Response {
         typed_buff.copy_to(&mut body);
         Ok(body)
     }
+
+    /// attempts to clone the request via the Response.clone api
+    pub fn try_clone(&self) -> Result<Self, Error> {
+        let clone = self.0.clone().map_err(js_to_error)?;
```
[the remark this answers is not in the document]
seanaye: This clone, is actually calling the MDN Response clone method, not the rust trait clone, this yields a proper clone in javascript

crates/net/src/http/headers.rs
```diff
@@ -6,6 +6,7 @@ use wasm_bindgen::{JsCast, UnwrapThrowExt};
 // I experimented with using `js_sys::Object` for the headers, since this object is marked
 // experimental in MDN. However it's in the fetch spec, and it's necessary for appending headers.
 /// A wrapper around `web_sys::Headers`.
+#[derive(Clone)]
```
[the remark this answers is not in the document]
seanaye: I fixed this by implementing the Clone trait manually

crates/net/src/http/headers.rs
```diff
@@ -37,14 +44,14 @@ impl Headers {
 
     /// This method appends a new value onto an existing header, or adds the header if it does not
     /// already exist.
-    pub fn append(&self, name: &str, value: &str) {
+    pub fn append(&mut self, name: &str, value: &str) {
```
hamza1311, line 47: I'm not a fan of making this `&mut`. This makes the method unnecessarily restrictive.

CC: @futursolo, do you have any thoughts on this?

crates/net/src/http/request.rs
```diff
@@ -265,12 +265,12 @@ impl Request {
     }
 
     /// Gets the body.
-    pub fn body(&self) -> Option<ReadableStream> {
+    pub fn body(self) -> Option<ReadableStream> {
```
hamza1311, line 268: This function can be called multiple times and still be valid.

crates/net/src/http/request.rs
```diff
@@ -279,12 +279,12 @@ impl Request {
...
     #[cfg(feature = "json")]
     #[cfg_attr(docsrs, doc(cfg(feature = "json")))]
-    pub async fn json<T: DeserializeOwned>(&self) -> Result<T, Error> {
+    pub async fn json<T: DeserializeOwned>(self) -> Result<T, Error> {
         serde_json::from_str::<T>(&self.text().await?).map_err(Error::from)
     }
 
     /// Reads the reqeust as a String.
-    pub async fn text(&self) -> Result<String, Error> {
+    pub async fn text(self) -> Result<String, Error> {
```
hamza1311, line 287: This change, while good in theory, also makes the API unnecessarily restrictive.

image

Notice how the object can still be used after `text()` is called

Alextopher:
We could also split bodies like Tower (I think it's Tower I'll double check later). But the idea is splitting the Response into (Body, Parts) where Parts has all of the normal components.

I'm not an expert rust user but it feels un-rusty to leave a class of bug the type system could easily solve.

hamza1311:
> I'm not an expert rust user but it feels un-rusty to leave a class of bug the type system could easily solve. 

I don't disagree. But at the same time, I also don't want to take away functionality that would otherwise be possible

databasedav:
can we just add a feature flag that always `.try_clone`'s and body when it's read? that way, we can preserve the references in the signatures, while avoiding the clone management stuff; or perhaps a separate auto-body-cloning struct?

PR #336, opened 2023-05-24
Add `PartialEq, Eq, PartialOrd, Ord, Hash` for eventsource `State`

Charles-Schleich:
The aim is to make it a little easier to debug and write more robust logic when handling an eventsource connection.

### DIFF
--- a/crates/net/src/eventsource/mod.rs
+++ b/crates/net/src/eventsource/mod.rs
@@ -11,7 +11,7 @@ use std::fmt;
 ///
 /// See [`EventSource.readyState` on MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventSource/readyState)
 /// to learn more.
-#[derive(Copy, Clone, Debug)]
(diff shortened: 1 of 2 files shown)


crates/net/src/eventsource/mod.rs
```diff
@@ -11,7 +11,7 @@ use std::fmt;
 ///
 /// See [`EventSource.readyState` on MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventSource/readyState)
 /// to learn more.
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
richardpringle, line 14: I'm not on the `gloo` team, but would it make sense to add `PartialEq, Eq, PartialOrd, Ord` behind a feature flag and `Hash` behind another? I think people care about code bloat when it comes to wasm (depending on where the code is being run).

hamza1311 (approved):

Issue #275, opened 2022-11-29
Send Blob to a Worker


hamza1311:
## Problem

The `gloo_worker::Worker` trait's `Input` type has a `serde::Serialize` trait bound. `web_sys` types, including `Blob`, do not implement this trait.

## Context

I have a `Worker` that  processes images. The input of it is a `gloo_file::File` (which internally holds a `Blob`). `File` cannot implement `Serialize`  because it's inner fields do not

futursolo:
You can use structured clone if you write a `gloo::worker::Codec` and implement the input and output type with `#[serde(skip)]` on the field with `web_sys::File`. You should be able to get around the Serialize and Deserialize requirement.

You can also read the image into a `Vec<u8>` before sending it to a Web Worker and send back as `Vic<u8>` or blob URL, as you have to convert the image into a base64 blob URL if you want to display or saving it anyways.

hamza1311:
> You can use structured clone if you write a `gloo::worker::Codec` and implement the input and output type with `#[serde(skip)]` on the field with `web_sys::File`. You should be able to get around the Serialize and Deserialize requirement with a custom Codec.

I don't quite follow here. Do you have any example that for this? 

> You can also read the image into a `Vec<u8>` before sending it to a Web Worker and send back as `Vec<u8>` or blob URL, as you have to convert the image into a base64 blob URL if you want to display or saving it anyway.

I was trying to avoid the cost of serializing `Vec<u8>` and sending the bytes. The image may be multiple MBs and I didn't want to hide a potential clone behind the serialization

yordanmadzhunkov:
I have a similar issue. I am trying to implement a yew front end application that uploads image file to a server. 
My back-end implementation supports multipart upload and I am able to upload images with curl from the command line.

I tried to serialize the image as array of u8 and whenever the post request size gets greater then 64kB, I run into some errors and I can't figure out how to continue forward.

It will be very helpful, if somebody provides with working example for this scenario using gloo

thepackett:
I am also needing to send a file to a Worker and can't figure out a solution, or understand what futursolo is suggesting. Processing large files can take several seconds, even minutes when you have large data sets to process, and freezing the entire webpage for that duration is extremely undesirable. Using a webworker would be the perfect solution, if only Blobs could be serialized and deserialized.

futursolo:
> take several seconds, even minutes when you have large data sets to process

[...]

PR #340, opened 2023-05-31
Add an example of processing transferrable types with worker

futursolo:
See: #275 

This pull request adds an example to demonstrate how to use custom codecs for transferrable types.

### DIFF
--- a/examples/file-hash/src/codec.rs
+++ b/examples/file-hash/src/codec.rs
@@ -25,21 +25,23 @@ impl TransferrableCodec {
     }
 }
 
+// This codec implementation relys on some internal implementation details about gloo worker message types.
+// This should be considered as a last resort approach after all other options are exhausted.
+//
+// In addition, using this approach could also potentially create unsound code.
+// This could cause data being sent to a wrong worker or with wrong handler id if it is not implemented
+// properly.
+//
+// You should only use this approach if you know how to make it sound.
+// This example mitigates this issue by only allowing 1 file to be processed at a time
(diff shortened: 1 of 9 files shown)


examples/file-hash/src/codec.rs
```diff
@@ -0,0 +1,88 @@
...
+// This codec implementation relys on some internal implementation details about gloo worker message types.
+// This should be considered as a last resort approach after all other options are exhausted.
+//
+// In addition, using this approach could also potentially create unsound code.
+// This could cause data being sent to a wrong worker or with wrong handler id if it is not implemented
+// properly.
+//
+// You should only use this approach if you know how to make it sound.
+// This example mitigates this issue by only allowing 1 file to be processed at a time
+// and only 1 worker instance across the entire tab.
```
hamza1311, line 37: This sounds like a pretty big gotcha, especially for an example. It _can_ work but I would argue this is broken.

The usecase I had for this required me to fork a worker from a svelte component and pass callback to the fork that mutates the component's state. I certainly don't see that working with this implementation

futursolo: I agree with you.

However, I do not have a better way of achieving it and I do not like the way I made it to work either.

The proper solution would be to wait until https://github.com/cloudflare/serde-wasm-bindgen/issues/32 resolved (there is a pending pull request), which brings `Serialize` and `Deserialize` to `JsValue`. The codec will work with transferrable types by simply using a codec implemented with `serde-wasm-bindgen` after that issue is resolved.

PR #347, opened 2023-06-02
doc: Inline documentation in the supercrate

oblique:
[gloo documentation](https://docs.rs/gloo/0.8.0/gloo/index.html) currently does not show you the documentation of the subcrates. Also searching does not work in the supercrate.

This can be solved by adding `#[doc(inline)]` on every crate reexport.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,9 +58,6 @@ history = ["gloo-history"]
 worker = ["gloo-worker"]
 net = ["gloo-net"]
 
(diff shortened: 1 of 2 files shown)


hamza1311:
#333 feature-gates the crates. Can you place `doc(cfg(_))` attributes to show they're feature gates?

oblique:
ready

hamza1311 (approved):
Thanks

PR #349, opened 2023-06-05
Add Worker Loader

futursolo:
This pull request implements the loader shim mode that is coming in the next trunk version.

This mode no longer uses inline scripts to load workers.
Trunk is responsible to create a loader script to initialise the worker.


futursolo:
Trunk v0.17 has been released so we can release this gloo worker mode as well.

hamza1311:
Can you please also update the chnagelog?

futursolo:
> Can you please also update the chnagelog?

This is done.

> Also, I would like to keep the ability to build workers without using trunk. That is kept intact, right? I use it to build workers that are called from a JS (SvelteKit) application so that is pretty important to have

This pull request adds a `spawn_with_loader()` function, which can be used with the loader shim created by trunk when `data-loader-shim` option is specified in `index.html`.

Existing worker users can continue using the `spawn()` function or create the loader shim manually.

e.g: The markdown example create the following loader shim:

```javascript
importScripts("./example_markdown_worker.js");wasm_bindgen("./example_markdown_worker_bg.wasm");
```

examples/markdown/src/bin/example_markdown_app.rs
```diff
@@ -22,7 +22,8 @@ fn main() {
         .flatten()
         .expect_throw("failed to query root element");
 
-    let mut bridge = MarkdownWorker::spawner().spawn("/example_markdown_worker.js");
+    let mut bridge =
+        MarkdownWorker::spawner().spawn_with_loader("/example_markdown_worker_loader.js");
```
hamza1311, line 26: Wonder if trunk can set a build-time env so this path doesn't need to be hard coded

hamza1311 (approved):
Thanks!

PR #357, opened 2023-07-23
history: Fix required feature set of serde dependency

jplatte:
The state module uses Serialize and Deserialize as derive macros, but those are only available with the derive feature.

Also simplify some cfg expressions in the same crate.


futursolo (approved):
Thanks for spotting this.

I think most of the time the derive feature is enabled by a downstream crate already, so we never caught this on our side.

PR #358, opened 2023-07-23
history: Use thread_local!'s lazy initialization instead of RefCell<Option<_>>

### LINKED ISSUE #356: Use OnceCell for gloo-history
The `BrowserHistory` initialization could be optimized / simplified a bit by using `OnceCell` instead of `RefCell<Option<_>>`. However, this would raise the MSRV to 1.70. Are you interested in a PR?

jplatte:
Closes #356.

### DIFF
--- a/crates/history/src/browser.rs
+++ b/crates/history/src/browser.rs
@@ -250,47 +250,34 @@ impl Default for BrowserHistory {
     fn default() -> Self {
         // We create browser history only once.
         thread_local! {
-            static BROWSER_HISTORY: RefCell<Option<BrowserHistory>> = RefCell::default();
-            static LISTENER: RefCell<Option<EventListener>> = RefCell::default();
+            static BROWSER_HISTORY: (BrowserHistory, EventListener) = {
+                let window = window();
+
+                let inner = window
+                    .history()
+                    .expect_throw("Failed to create browser history. Are you using a browser?");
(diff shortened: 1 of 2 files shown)


hamza1311:
I'm fine with the change but MSRV bump makes me not want to make this change. It means that not only gloo but also libraries that depend on gloo have to bump their MSRV. 

Can we use the once_cell crate if OnceCell is not available from std?

jplatte:
Sure, but how should Cargo decide whether to pull in the dependency or not? If I make it an optional dependency, this is still an MSRV bump change for people who were using `default-features = false`. Is that acceptable? We could also pull in `once_cell` unconditionally, it's a crate that very oftens ends up in people's dep tree anyways.

futursolo:
How about we use once_cell::unsync::Lazy for all versions until it becomes available in stable? This avoids a MSRV bump and promotes lazy initialisation, which fits this use case better.

I do not feel comfortable that we should bump MSRV to the latest stable given latest stable has issues with WASM-bindgen (https://github.com/rust-lang/rust/issues/111888).

jplatte:
I looked at the docs for `thread_local!` more closely and found that it supports lazy initialization all on its own. Have updated the PR with an implementation that's simpler than everything we have discussed before :)

hamza1311 (approved):
Thanks!

PR #360, opened 2023-07-29
Migrate to Edition 2021

futursolo:
This pull request migrates to edition 2021.

Since wasm-bindgen requires 1.56, there is no benefit to keep crates at edition 2018.
Upgrading the crates to edition 2021 will enable the rust-version field which suppresses clippy for lints not available in MSRV.

I have calculated the MSRV with `cargo-msrv`, which is 1.64.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,13 +7,17 @@ on:
 
 jobs:
   native_tests:
+    strategy:
+      fail-fast: false
+      matrix:
+        rust-version: [1.64, stable, nightly]
     name: Native Tests
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: dtolnay/rust-toolchain@master
         with:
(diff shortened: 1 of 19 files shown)


futursolo:
Tests are failing as Trunk released a new version 10 minutes ago.

hamza1311:
Can you please also update the changelog?

Also, I believe this should be a minor version, not patch, when releasing, right?

futursolo:
> Can you please also update the changelog?

This is done. I think we should have a PR template so people don't forget about this.

> Also, I believe this should be a minor version, not patch, when releasing, right?

I believe so as this pull request now requires 1.64 for all crates where as in the past only some crates require 1.64.

hamza1311:
> This is done. Should we add a PR template so future contributors don't forget about this?

Good idea!

I also updated the CI so it runs against MSRV (and nightly) too. Will merge this once CI is green

PR #363, opened 2023-08-06
Add traits PartialEq and Eq to EventListenerOptions, and EventListenerPhase

Philipp-M:
I needed to compare the options of an event listener, I don't think it hurts adding support for `PartialEq` and `Eq` for the relevant types.


hamza1311 (changes requested):
Thanks. Can you please also update CHANGELOG.md at the root of the repo?

Philipp-M:
Like this?

Issue #362, opened 2023-08-05
gloo-histroy Support custom query decoder / encoder


Roba1993:
## Summary

Allow custom query decoder / encoder to support advanced use-cases.

## Motivation

WI have the problem, that I need to serialize and deserialize rust structs into the a query. This is done via serde and serde_json. For better readability and changeability of the url I want to use an own serializer / deserializer. In my example along the lines of [jsonurl.org](https://jsonurl.org).

## Detailed Explanation

Unfortunately I have not a clear idea of to implement this right now in gloo.

## Drawbacks, Rationale, and Alternatives

Drawback could be a more complex API? 

An alternative would also to implement a [jsonurl.org](https://jsonurl.org) behavior natively into gloo-history. But this is would not be completely HTML conform and would break existing links.

## Unresolved Questions

I also tried to use jsonurl to create a query string and push that as query. But it's still getting serialized / deserialized with the default behavior...

hamza1311:
What prevents you from pushing the query as:
```rust
struct Query {
    json_url: String
}
```
The url will be `?json_url=<query>`

xfbs:
Hey, I just now also have the same issue. I have a Query struct that is really simple, like this:

```rust
#[derive(Serialize, Deserialize)]
struct Query {
    pub query: Vec<Predicate>,
    pub sort: Option<String>,
}
```

And I get an error. 

I believe that the underlying issue for this is that `gloo` uses `serde_urlencoded` to do the query string encoding and decoding, which is quite limited in terms of what it supports. 

It seems that the `serde_qs` crate is a more powerful and actively maintained implementation of query string encoding and decoding. It supports any arbitrary Rust structs, enums and basically anything the type system allows.

I already use this crate to decode my querys with axum. For that reason, the crate has an `axum` feature. When enabled, it offers the `serde_qs::axum::QsQuery` extractor that has an interface which matches that of the built-in extractor.

I would recommend one of two options:

- adding a feature flag to `gloo-history`, maybe named `query-qs` which can be enabled to use `serde-qs` instead.
- making the interface accept an enum
- adding separate, optional methods to use a different encoder

I think this would bring quite some value to people using more advanced query parameters. However, this is a breaking change I believe. 

If you would like, I could draft up a PR with my idea for how it could look like, in a way that does not break existing code.

Cheers and thanks for the work on this crate!

Roba1993:
@hamza1311 the answer from @xfbs is really on the point.
In addition to your question:
[...]

PR #364, opened 2023-08-08
Introduces the FromQuery and IntoQuery traits.

xfbs:
Introduces the `FromQuery` and `IntoQuery` traits to allow for customizing how query strings are encoded and decoded. Introduces the

- This PR is related to issue #362.
- I have taken care to make sure that this PR does not break any consumers of this crate. 

Because this PR introduces quite a few new things, allow me to walk through it an explain what is happening here.

### Previous Behaviour

The `History` trait has some methods that allow specifying a query parameter. This query parameter must implement `Serialize`. It is encoded using `serde_urlencoded`. These return a `HistoryResult<()>`, which expands to a `Result<(), HistoryError>`.

```rust
impl History {
    fn push_with_query<'a, Q>(&self, route: impl Into<Cow<'a, str>>, query: Q) -> HistoryResult<()>
        where Q: Serialize;
    fn replace_with_query<'a, Q>(&self, route: impl Into<Cow<'a, str>>, query: Q) -> HistoryResult<()>
        where Q: Serialize;
    ...
}
```

### New Behaviour

We introduce a helper trait, `IntoQuery`, to help us encode any arbitrary type that implements this into a query string. It has a configurable error (because different strategies might fail in different ways) and a method to encode that takes `&self`.

```rust
pub trait IntoQuery {
    type Error;
    fn encode(&self) -> Result<Cow<'_, str>, Self::Error>;
}
```

We also add a blanket implementation of this type for any type that implements `Serialize` with the `Error` set to `HistoryError`.

```rust
impl<T: Serialize> IntoQuery for T {
    type Error = HistoryError;

    fn encode(&self) -> Result<Cow<'_, str>, Self::Error> {
        serde_urlencoded::to_string(self)
            .map(Into::into)
            .map_err(Into::into)
    }
}
```

Now, we change the methods in the `History` trait to constrain on anything that implements `IntoQuery` instead of `Serialize`. Since we have a blanket implementation of `IntoQuery` for anything that implements `Serialize`, these two constraints are identical at this point. These now return a `IntoQuery::Error` as an error instead.

```rust
impl History {
    fn push_with_query<'a, Q>(&self, route: impl Into<Cow<'a, str>>, query: Q) -> HistoryResult<(),Q::Error>
        where Q: IntoQuery;
    fn replace_with_query<'a, Q>(&self, route: impl Into<Cow<'a, str>>, query: Q) -> HistoryResult<(), Q::Error>
        where Q: IntoQuery;
    ...
}
```

Before, it was only possible to call these methods with `Q: Serialize` and it returned a `Result<(), HistoryError>`. Now, calling it with any type that is `Q: Serialize` is still possible, and returns the same value. However, it is now possible to implement `IntoQuery` manually to override an encoding scheme. This is done with the `Qs` type and the `Raw` type.

### Breaking Changes

Because the new `History` trait API is a superset of the previous API, this does not constitute a breaking change for consumers of this trait. It does however change the trait, which means that anyone who manually implements the `History` trait for a custom type (I don't know why anyone would do that, but you could) needs to adjust their implementation slightly. It might be worth looking into [sealing the History trait](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/) for future-proofing it, otherwise any change, even addition of a method, is a potential (hypothetical) breaking change.

### Previous Behaviour

Query decoding happens in the `Location` struct, which has a `query` method.

```rust
impl Location {
        pub fn query<T>(&self) -> HistoryResult<T> where T: DeserializeOwned;
}
```

### New Behaviour

We introduce a trait to capture the idea of decoding a query string. This has a `Target` type parameter, so that a `MyCustomStrategy` can decode into a `String` if it wanted to, and an `Error` parameter, as well as a `decode()` method.

```rust
pub trait FromQuery {
    type Target;
    type Error;
    fn decode(query: &str) -> Result<Self::Target, Self::Error>;
}
```

Again, we add a blanket implementation for any T that implements `DeserializeOwned`.

```rust
impl<T: DeserializeOwned> FromQuery for T {
    type Target = T;
    type Error = HistoryError;

    fn decode(query: &str) -> Result<Self::Target, Self::Error> {
        serde_urlencoded::from_str(query).map_err(Into::into)
    }
}
```

With this, we change the method to return a `T::Target` and an `T::Error` instead.

```rust
impl Location {
pub fn query<T>(&self) -> HistoryResult<T::Target, T::Error> where T: FromQuery;
```

With this change, all previous invocations still work the exact same way. However, we can now implement `FromQuery` for custom types to use a custom decoding strategy.

### Breaking Changes

Since this is not a trait, and the new definition is a superset of the old, this does not constitute a breaking change.

## Tests

Tests have been added for the traits. The test coverage exceeds 80%.

## Documentation

Documentation has been added, however it could still be improved with more examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ### Next Version
 
 - Migrate to Edition 2021 and Apply MSRV in Cargo.toml (#360)
+- Introduces the `FromQuery` and `ToQuery` traits to allow for customizing
+  how query strings are encoded and decoded in `gloo_history`. (#364)
 
 ### Version "0.2.3"
 
--- a/crates/history/src/any.rs
+++ b/crates/history/src/any.rs
@@ -1,16 +1,13 @@
 use std::borrow::Cow;
 
-#[cfg(feature = "query")]
-use serde::Serialize;
-
 use crate::browser::BrowserHistory;
-#[cfg(feature = "query")]
-use crate::error::HistoryResult;
 use crate::hash::HashHistory;
 use crate::history::History;
 use crate::listener::HistoryListener;
 use crate::location::Location;
 use crate::memory::MemoryHistory;
+#[cfg(feature = "query")]
+use crate::{error::HistoryResult, query::ToQuery};
 
 /// A [`History`] that provides a universal API to the underlying history type.
 #[derive(Clone, PartialEq, Debug)]
@@ -79,9 +76,13 @@ impl History for AnyHistory {
     }
 
     #[cfg(feature = "query")]
-    fn push_with_query<'a, Q>(&self, route: impl Into<Cow<'a, str>>, query: Q) -> HistoryResult<()>
+    fn push_with_query<'a, Q>(
+        &self,
+        route: impl Into<Cow<'a, str>>,
+        query: Q,
+    ) -> HistoryResult<(), Q::Error>
     where
-        Q: Serialize,
+        Q: ToQuery,
     {
         match self {
             Self::Browser(m) => m.push_with_query(route, query),
@@ -94,9 +95,9 @@ impl History for AnyHistory {
         &self,
         route: impl Into<Cow<'a, str>>,
         query: Q,
-    ) -> HistoryResult<()>
+    ) -> HistoryResult<(), Q::Error>
     where
-        Q: Serialize,
+        Q: ToQuery,
     {
         match self {
             Self::Browser(m) => m.replace_with_query(route, query),
@@ -111,9 +112,9 @@ impl History for AnyHistory {
         route: impl Into<Cow<'a, str>>,
         query: Q,
         state: T,
-    ) -> HistoryResult<()>
+    ) -> HistoryResult<(), Q::Error>
     where
-        Q: Serialize,
+        Q: ToQuery,
         T: 'static,
     {
         match self {
@@ -129,9 +130,9 @@ impl History for AnyHistory {
         route: impl Into<Cow<'a, str>>,
         query: Q,
         state: T,
-    ) -> HistoryResult<()>
+    ) -> HistoryResult<(), Q::Error>
     where
-        Q: Serialize,
+        Q: ToQuery,
         T: 'static,
     {
         match self {
--- a/crates/history/src/query.rs
+++ b/crates/history/src/query.rs
@@ -1,0 +1,112 @@
+//! # Encoding and decoding strategies for query strings.
+//!
+//! There are various strategies to map Rust types into HTTP query strings. The [`FromQuery`] and
+//! [`ToQuery`] encode the logic for how this encoding and decoding is performed. These traits
+//! are public as a form of dependency inversion, so that you can override the decoding and
+//! encoding strategy being used.
+//!
+//! These traits are used by the [`History`](crate::History) trait, which allows for modifying the
+//! history state, and the [`Location`](crate::Location) struct, which allows for extracting the
+//! current location (and this query).
+//!
+//! ## Default Strategy
+//!
+//! By default, any Rust type that implements [`Serialize`] or [`Deserialize`](serde::Deserialize)
+//! has an implementation of [`ToQuery`] or [`FromQuery`], respectively. This implementation uses
+//! the `serde_urlencoded` crate, which implements a standards-compliant `x-www-form-urlencoded`
+//! encoder and decoder. Some patterns are not supported by this crate, for example it is not
+//! possible to serialize arrays at the moment. If this is an issue for you, consider using the
+//! `serde_qs` crate.
+//!
+//! Example:
+//!
+//! ```rust,no_run
+//! use serde::{Serialize, Deserialize};
+//! use gloo_history::{MemoryHistory, History};
+//!
+//! #[derive(Serialize)]
+//! struct Query {
+//!     name: String,
+//! }
+//!
+//! let query = Query {
+//!     name: "user".into(),
+//! };
+//!
+//! let history = MemoryHistory::new();
+//! history.push_with_query("index.html", &query).unwrap();
+//! ```
+//!
+//! ## Custom Strategy
+//!
+//! If desired, the [`FromQuery`] and [`ToQuery`] traits can also be manually implemented on
+//! types to customize the encoding and decoding strategies. See the documentation for these traits
+//! for more detail on how this can be done.
+use crate::error::HistoryError;
+use serde::{de::DeserializeOwned, Serialize};
+use std::borrow::Cow;
+use std::convert::{AsRef, Infallible};
+
+/// Type that can be encoded into a query string.
+pub trait ToQuery {
+    /// Error that can be returned from the conversion.
+    type Error;
+
+    /// Method to encode the query into a string.
+    fn to_query(&self) -> Result<Cow<'_, str>, Self::Error>;
+}
+
+/// Type that can be decoded from a query string.
+pub trait FromQuery {
+    /// Target type after parsing.
+    type Target;
+    /// Error that can occur while parsing.
+    type Error;
+
+    /// Decode this query string into the target type.
+    fn from_query(query: &str) -> Result<Self::Target, Self::Error>;
+}
+
+impl<T: Serialize> ToQuery for T {
+    type Error = HistoryError;
+
+    fn to_query(&self) -> Result<Cow<'_, str>, Self::Error> {
+        serde_urlencoded::to_string(self)
+            .map(Into::into)
+            .map_err(Into::into)
+    }
+}
+
+impl<T: DeserializeOwned> FromQuery for T {
+    type Target = T;
+    type Error = HistoryError;
+
+    fn from_query(query: &str) -> Result<Self::Target, Self::Error> {
+        serde_urlencoded::from_str(query).map_err(Into::into)
+    }
+}
+
+/// # Encoding for raw query strings.
+///
+/// The [`Raw`] wrapper allows for specifying a query string directly, bypassing the encoding. If
+/// you use this strategy, you need to take care to escape characters that are not allowed to
+/// appear in query strings yourself.
+#[derive(Debug, Clone)]
+pub struct Raw<T>(pub T);
+
+impl<T: AsRef<str>> ToQuery for Raw<T> {
+    type Error = Infallible;
+
+    fn to_query(&self) -> Result<Cow<'_, str>, Self::Error> {
+        Ok(self.0.as_ref().into())
+    }
+}
+
+impl<T: for<'a> From<&'a str>> FromQuery for Raw<T> {
+    type Target = T;
+    type Error = Infallible;
+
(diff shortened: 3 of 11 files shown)


crates/history/Cargo.toml
```diff
@@ -30,4 +31,5 @@ gloo-timers = { version = "0.2.6", features = ["futures"], path = "../timers" }
 
 [features]
 query = ["thiserror", "serde_urlencoded"]
+query-qs = ["query", "dep:serde_qs"]
```
futursolo, line 34: I am not in favour of including additional serialisers / deserialisers if a decoding / encoding trait can be implemented from outside of `gloo-history`.

xfbs: I am happy to remove it, there is precedent for adding integrations into `serde-qs` itself. 

Although, personally I would advocate for leaving it in for a batteries-included experience. I feel that a lot of people will hit a wall with the default due to [serde_urlencoded not wanting to support sequences](https://github.com/nox/serde_urlencoded/issues/109). 

Even the [urlencoded spec](https://url.spec.whatwg.org/#urlencoded-parsing) supports sequences. They are generated by standards-compliant browsers in HTTP forms linked in the issue. The failure to support these seriously limits the usefulness of the query encoding and decoding for anything but the most basic applications. By shipping with an optional, but more complete strategy, at least we have something to point people at when they run into issues.

futursolo: The primary concern is that if someone else in the future wants a third way of query string especially when one is RFC'ed.
I do not feel there is any more good ground of not accepting such a pull request if we start allowing additional formats in the crate itself.

> Even the [urlencoded spec](https://url.spec.whatwg.org/#urlencoded-parsing) supports sequences. They are generated by standards-compliant browsers in HTTP forms linked in the issue. The failure to support these seriously limits the usefulness of the query encoding and decoding for anything but the most basic applications. 

In `serde_urlencoded`, a sequence can be defined as `Vec<(String, String)>` for arbitrary valid x-form-www-urlencoded sequence, should you need it. `serde_qs` simply do not have a way to produce a sequence without `[]` suffix.

However, because `serde_qs` interprets this differently than the browsers itself, this also limits the usefulness of the implementation.

```html
<form>
    <input name="tag">
    <input name="tag">
    <input type="submit" value="Submit">
</form>
```

Browsers will convert this to `tag=biking&tag=shedding`, however, if I try to use serde_qs to parse the value, it will say: `Custom("Multiple values for one key: \"tag\"")`.

xfbs: It is tricky indeed. 

I understand the possibility to use a `Vec<(String, String)>`, but does that not mean you lose all of the type-safety benefits? Because now you have to manually pull out what key goes where in your struct? 

Possibly an idea would be to go over to that crate and see if we can convince @nox to support sequences, even if it is behind a feature flag. In all of the closed issues, he has indicated that he prefers people use `serde_qs` due to him not wanting to have allocations in the crate (hence why I suggested we build this in here).

 I think I do understand his thinking, but I am quite concerned about having foundational crates like `gloo` or `axum` where basic stuff such as query sequences do not work properly out of the box. It really threw me off when I found out this not to be working in `axum`, and had to awkwardly code around it :/

Will remove the support for `serde_qs` for now so we can have this move ahead. But I will also take a look into the `serde_urlencoded` crate and see if there is something I can do on that side to unbreak this and unblock the people running into this issue.

crates/history/src/query.rs
```diff
@@ -0,0 +1,120 @@
...
+
+/// Type that can be decoded from a query string.
+pub trait FromQuery {
+    /// Target type after parsing.
+    type Target;
+    /// Error that can occur while parsing.
+    type Error;
+
+    /// Decode this query string into the target type.
+    fn decode(query: &str) -> Result<Self::Target, Self::Error>;
```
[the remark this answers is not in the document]
xfbs: I think the issue with `into_query` is that as you mentioned earlier, `Into` usually suggests something that cannot fail. Therefore it would need to be named `try_into_query()` and `try_from_query()`. But that point, I would rather suggest to rename the trait into `DecodeQuery` and `EncodeQuery`, because semantially the idea of decoding and encoding already suggests that it is an operation which can fail.

What do you think about that idea?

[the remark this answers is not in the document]
futursolo: I am fine with both solutions as `from` / `to` is not strictly interpreted as infallible in std. E.g.: `FromStr::from_str` is defined as `fn from_str(s: &str) -> Result<Self, Self::Err>;` and `OsStr::to_str` is defined as `pub fn to_str(&self) -> Option<&str>`.

> But that point, I would rather suggest to rename the trait into DecodeQuery and EncodeQuery, because semantially the idea of decoding and encoding already suggests that it is an operation which can fail.

It might make sense to combine decode and encode side of queries into `Query` trait as a query type in this case usually is used as a type to decode / encode the actual underlying query so it should implement both operations.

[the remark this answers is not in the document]
xfbs: I have taken both of your suggestions and renamed the traits to `FromQuery`/`ToQuery` and the methods to `from_query` and `to_query`.

> It might make sense to combine decode and encode side of queries into Query trait as a query type in this case usually is used as a type to decode / encode the actual underlying query so it should implement both operations.

Well, the two operations are different (encoding vs decoding). We do not want to force users of the API to implement both just because they want to use the `push_with_query()` API. That would not be a good design.

We can always have a supertrait:

```rust
pub trait Query {
    type Decoded;
    type DecodeError;
    type EncodeError;
    fn from_query()
    fn to_query()
}

impl<T: FromQuery + ToQuery> Query for T {
    type Decoded = <T as FromQuery>::Target;
    // ...
}
```

But we do not have a use for this type, as we have no method that requires **both** the ability to encode and to decode. All the APIs require either one but not both. That is why I think it is good that these are separate traits.

crates/history/src/any.rs
```diff
@@ -1,16 +1,12 @@
...
-use crate::error::HistoryResult;
-use crate::hash::HashHistory;
-use crate::history::History;
-use crate::listener::HistoryListener;
-use crate::location::Location;
-use crate::memory::MemoryHistory;
+use crate::{
+    browser::BrowserHistory, hash::HashHistory, history::History, listener::HistoryListener,
+    location::Location, memory::MemoryHistory,
+};
```
hamza1311, line 9: Nit: this (and other similar import changes) are needed changes that appears in a diff. Do you have your editor set to forcefully follow this formatting style, perhaps? Rustfmt by default doesn't do this.

If you don't want to revert it, that is fine too. I realized this gloo doesn't have a `rustfmt.toml`. I'll add that after this PR, to avoid introducing conflicts

crates/history/src/query.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Method to encode the query into a string.
+    fn to_query(&self) -> Result<Cow<'_, str>, Self::Error>;
+}
+
+/// Type that can be decoded from a query string.
+pub trait FromQuery {
+    /// Target type after parsing.
+    type Target;
+    /// Error that can occur while parsing.
+    type Error;
```
hamza1311, line 66: Is there a reason why `Error` doesn't have a [`std::error::Error`](https://doc.rust-lang.org/nightly/core/error/trait.Error.html) trait bound?

crates/history/tests/query.rs
```diff
@@ -0,0 +1,52 @@
+#![cfg(feature = "query")]
```
hamza1311, line 1: I would also run these tests inside a browser:
```rust
#[cfg(target_arch = "wasm32")]
use wasm_bindgen_test::{wasm_bindgen_test as test, wasm_bindgen_test_configure};
#[cfg(target_arch = "wasm32")]
wasm_bindgen_test_configure!(run_in_browser);
```

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ### Next Version
 
 - Migrate to Edition 2021 and Apply MSRV in Cargo.toml (#360)
+- Introduces the `ToQuery` and `IntoQuery` traits to allow for customizing
```
hamza1311, line 6:
```suggestion
- Introduces the `FromQuery` and `ToQuery` traits to allow for customizing
```

hamza1311:
I wonder if there is a way to use `URLSearchParams` with serde

crates/history/src/any.rs
xfbs: I may have ran some keybinding that runst rustfmt with overriden defaults -- will revert this!

crates/history/src/query.rs
xfbs: I decided not to bound it here because we don't do anything with the error besides return it. If you were using this trait anywhere and you do care that the error type implements `std::error::Error`, you can always do:

```rust
fn do_something<T: FromQuery>() where T::Error: std::error::Error {}
```

If you use the APIs with anything that implements `Serialize` then the error type will be serialize to. So if you bound is `T: Serialize`, then it works as expected even without the `where` bound.

My rule of thumb is to keep traits as generic as possible — even if 99% of users would have an Error type that implements `std::error::Error`, unless I use that directly, I do not constrain it. One example is `anyhow::Error`, which does not implement `std::error::Error`  (for technical reasons) and yet is widely used. So adding a constraint of `type Error: std::error::Error` would make it impossible to use `anyhow::Error` as the error type (iirc).

crates/history/tests/query.rs
xfbs: Neat! I was able to run the headless browser tests locally with this.

xfbs:
> I wonder if there is a way to use `URLSearchParams` with serde

Do you mean something like extracting key-value pairs from a `URLSearchParams` and feeding that into a serde deserializer?

hamza1311:
> > I wonder if there is a way to use `URLSearchParams` with serde
> 
> Do you mean something like extracting key-value pairs from a `URLSearchParams` and feeding that into a serde deserializer? 

I mean the opposite of that: using serde to serialize into a URLSearchParams

xfbs:
That should be possible, actually. Although, you would probably not want to couple it so strongly. So ideally, you'd want something that serializes into a `Vec<(String, String)>`, and then you feed that into a `URLSearchParams`. Sounds like a fun weekend project.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ### Next Version
 
 - Migrate to Edition 2021 and Apply MSRV in Cargo.toml (#360)
+- Introduces the `FromQuery` and `ToQuery` traits to allow for customizing
+  how query strings are encoded and decoded in `gloo_history`.
```
futursolo, line 7:
```suggestion
  how query strings are encoded and decoded in `gloo_history`. (#364)
```

Could you please include the pull request ID as other entries in the changelog?

futursolo (approved):
Thanks! LGTM.

I am approving this pull request as I do not have more comments with the current implementation and we can proceed with it after the change log is updated.

hamza1311 (approved):
Thanks for making this change!

PR #365, opened 2023-08-10
Add `TryFrom<web_sys::WebSocket>`

mokeyish:
On server side, websocket may create from WebSocketPair. So we need `TryFrom<web_sys::WebSocket>`.

https://docs.rs/worker/latest/src/worker/websocket.rs.html#26-31

### DIFF
--- a/crates/net/src/websocket/futures.rs
+++ b/crates/net/src/websocket/futures.rs
@@ -238,6 +238,20 @@ impl WebSocket {
     }
 }
 
+impl TryFrom<web_sys::WebSocket> for WebSocket {
+    type Error = JsError;
+
+    fn try_from(ws: web_sys::WebSocket) -> Result<Self, Self::Error> {
+        Self::setup(Ok(ws))
+    }
+}
+
+impl AsRef<web_sys::WebSocket> for WebSocket {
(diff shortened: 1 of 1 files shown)


crates/net/src/websocket/futures.rs
```diff
@@ -238,6 +238,20 @@ impl WebSocket {
...
+        Self::setup(Ok(ws))
+    }
+}
+
+impl AsRef<web_sys::WebSocket> for WebSocket {
+    fn as_ref(&self) -> &web_sys::WebSocket {
+        &self.ws
+    }
+}
+
```
hamza1311, line 254: Please remove this impl. We don't want to allow such easy access to the inner WebSocket instance. See the discussion at https://github.com/rustwasm/gloo/pull/337 for more details about why that is the case.

hamza1311 (changes requested):
Thanks for the PR! I'm in favor of adding a way to use an existing web socket instance. There's one comment that about accessing the inner instance that need to be addressed

hamza1311 (approved):
Thanks

carlsverre:
@hamza1311 is it possible to ship this sometime soon? I would prefer not to pin a git sha

Issue #25, opened 2019-03-15
How about impl AsyncRead/AsyncWrite for WebSocket?


najamelan:
### Motivation

Would allow frontend and backend code to communicate in an abstract manner on an async Stream. It allows code reuse and prevents every application to have to implement this on top of the existing websocket api

### Proposed Solution

Add impl AsyncRead/AsyncWrite to websocket

### Alternatives

Allow upgrading an Http connection to an AsyncRead/AsyncWrite without websockets. I have no idea what the implementation of this would look like, and if it's even possible without websockets.

That being said, this alternative would be preferable I think. It means we don't need WS on the server either, just an upgraded http connection. 

### Additional Context

I kind of need this. If need be I don't mind attempting to implement this myself. If this is deemed a good solution, is there any mentoring possible to hint the way?

alexcrichton:
Thanks for the report! I've transferred this over to the `gloo` repository here from the `wasm-bindgen` repository becaues the `web_sys` crate is intended to be a low level binding to web APIs, but this would be perfect for the `gloo` project!`

najamelan:
@alexcrichton  ok, that's cool. Currently what I'm looking at is using the `[set_onmessage](https://docs.rs/web-sys/0.3.16/web_sys/struct.WebSocket.html#method.set_onmessage)` on a websocket. If I understand well, the way to do this is by using a `wasm-bindgen::Closure` to create the callback Function. 

I would have to expose something that has `connect( url )` and returns an async `Stream`... Does that sound like the right approach to you? Does that seem realistic or do you see some technical blockers for this?

I also wonder, if I understood well, data from javascript has to be copied into the wasm memory. Is there a way to avoid this extra copy on the data that comes in from the websocket?

fitzgen:
See also https://github.com/rustwasm/gloo/issues/7

najamelan:
I'm making good progress on this I think, but I need a bit of advice. There isn't all that much example code or docs for this stuff. I can build streams that yield MessageEvent, but then the only way I see from the docs to get the data out of there and in rust is by using into_serde. This will use json stringify... I doubt that's what we want. There seem to be ways to share memory pointers from javascript like with WebAssembly.Memory and such, but what is the best way to get a bytearray out of a websocket connection into rust with wasm-bindgen, wasm-pack? Do I just write custom javascript for this? Given that this is a rust library, how do we ship the js?

najamelan:
[...]

PR #379, opened 2023-10-04
feat(net): AsyncWrite and AsyncRead for WebSocket

CBenoit:
Implements `AsyncWrite` and `AsyncRead` on `WebSocket` behind the `io-util` feature flag.

This code was originally living in an ad hoc fashion, in our [IronRDP repository](https://github.com/Devolutions/IronRDP/blob/3d98cd1d9476e1fae84f21771df2ef7f4cd189f7/crates/ironrdp-web/src/websocket.rs), and I thought it was best to upstream it. In addition to the unit tests, I can confirm it works when running long-lived RDP sessions in the browser.

Closes https://github.com/rustwasm/gloo/issues/25

### DIFF
--- a/crates/net/Cargo.toml
+++ b/crates/net/Cargo.toml
@@ -27,6 +27,7 @@ gloo-utils = { version = "0.2", path = "../utils", default-features = false }
 wasm-bindgen-futures = "0.4"
 futures-core = { version = "0.3", optional = true }
 futures-sink = { version = "0.3", optional = true }
+futures-io = { version = "0.3", optional = true }
 
 thiserror = "1.0"
 
@@ -45,7 +46,7 @@ serde = { version = "1.0", features = ["derive"] }
 once_cell = "1"
 
 [features]
-default = ["json", "websocket", "http", "eventsource"]
+default = ["json", "websocket", "http", "eventsource", "io-util"]
 
 # Enables `.json()` on `Response`
 json = ["serde", "serde_json", "gloo-utils/serde"]
@@ -99,3 +100,5 @@ eventsource = [
     'web-sys/EventSource',
     'web-sys/MessageEvent',
 ]
+# As of now, only implements `AsyncRead` and `AsyncWrite` on `WebSocket`
+io-util = ["futures-io"]
--- a/crates/net/Cargo.toml
+++ b/crates/net/Cargo.toml
@@ -46,7 +46,7 @@ serde = { version = "1.0", features = ["derive"] }
 once_cell = "1"
 
 [features]
-default = ["json", "websocket", "http", "eventsource", "io-util"]
+default = ["json", "websocket", "http", "eventsource"]
 
 # Enables `.json()` on `Response`
 json = ["serde", "serde_json", "gloo-utils/serde"]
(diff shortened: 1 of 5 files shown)


crates/net/Cargo.toml
```diff
@@ -99,3 +100,5 @@ eventsource = [
     'web-sys/EventSource',
     'web-sys/MessageEvent',
 ]
+# As of now, only implements `AsyncRead` and `AsyncWrite` on `WebSocket`
+io-util = ["futures-io"]
```
CBenoit, line 104: I figured it was best to provide this as an optional feature so that `futures-io` is not added in the dependency graph of all dependents

crates/net/src/websocket/futures.rs
```diff
@@ -57,6 +57,11 @@ pub struct WebSocket {
         Closure<dyn FnMut(web_sys::Event)>,
         Closure<dyn FnMut(web_sys::CloseEvent)>,
     ),
+    /// Leftover bytes when using `AsyncRead`.
+    ///
+    /// These bytes are drained and returned in subsequent calls to `poll_read`.
+    #[cfg(feature = "io-util")]
+    pub(super) read_pending_bytes: Option<Vec<u8>>, // Same size as `Vec<u8>` alone thanks to niche optimization
```
CBenoit, line 64: An alternative to `Vec::drain` would be to maintain an index pointing to the beginning of the "pending" region

crates/net/Cargo.toml
hamza1311, line 104:
```suggestion
io-util = ["dep:futures-io"]
```
This ensures `futures-io` is not a feature (by default, cargo adds features for optional dependencies)

hamza1311 (changes requested):
Can you also add an integration test for this and also update the CHANGELOG file in the root of the repo?

crates/net/Cargo.toml
CBenoit: I agree with you. However, the rest of the Cargo.toml is currently not using `dep:` to denote a dependency. When using this notation at one place, [the other places must be updated too because this effectively disables the old "implicit features"](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies). I think it’s better to tackle this in a separate PR for all crates in order to keep consistency. However, if you prefer I’m okay with converting this crate as part of this PR. I just want to make sure it’s okay for you.

CBenoit:
> Can you also add an integration test for this and also update the CHANGELOG file in the root of the repo?
> 
> Update: I just noticed that the tests mod includes the integration tests. Can you just update the changelog? It'll be ready for merge then

Maybe it’s better if I move these tests in the tests/ folder instead? I mirrored how it was done in  the `futures` module, but as you pointed it, those are integration tests. Maybe I should move the one in `futures` as well for consistency. What do you think?

hamza1311:
No, it's fine here as well. I would like to keep them consistent and this PR is not a place to move the tests. Can you merge from master and update the changelog? Merging from master should also fix the CI failures

CBenoit:
> No, it's fine here as well. I would like to keep them consistent and this PR is not a place to move the tests. Can you merge from master and update the changelog? Merging from master should also fix the CI failures

Sounds good. I rebased on master and only updated the changelog (see the last commit).

hamza1311 (approved):
Thank you very much!

PR #388, opened 2023-10-07
Remove Cloning on WorkerBridge

### LINKED ISSUE #387: Cloning gloo-worker bridges does not assign new HandlerIds
### Steps to Reproduce

Discovered in https://github.com/yewstack/yew/pull/3435.

Create a `WorkerBridge`, clone 5 times, then drop all of them.
You will see 1 connected message but 5 disconnected messages for the same `HandlerId`, which is not correct.

If applicable, add a link to a test case (as a zip file or link to a repository
we can clone).

### Expected Behavior

Bridge should either not be clonable, or assign unique `HandlerId` when being cloned, or only send disconnect message when all cloned copies of a worker bridge gets dropped.

### Actual Behavior

A clear and concise description of what actually happened.

A disconnect message is sent for the number of clones a worker bridge had.

### Additional Context

N/A

futursolo:
Fix #387.

This pull request removes cloning of bridges.
This is a breaking change.

If users want to share worker bridges, they can use `Rc`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -607,6 +607,7 @@ dependencies = [
  "futures",
  "futures-channel",
  "futures-core",
+ "futures-io",
  "futures-sink",
  "gloo-utils 0.2.0",
  "http",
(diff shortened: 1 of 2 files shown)


hamza1311 (approved):

hamza1311:
Released in gloo-worker v0.4.1 :tada:

PR #405, opened 2023-11-20
Add WASI support for gloo-history.

langyo:
I've been trying to use `yew` to render the page into the static HTML string on WASI. However, `gloo-history` cannot distinguish the browser WASM target (`wasm32-unknown-unknown` with `wasm-bindgen`) and WASI target (`wasm32-wasi`), and it would choose wrong functions for `wasm32-*`.

I have fixed it by using `target_os`, and it would fix the problem.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,6 @@ utils = ["gloo-utils"]
 history = ["gloo-history"]
 worker = ["gloo-worker"]
 net = ["gloo-net"]
-wasi = ["gloo-history/wasi"]
 
 [workspace]
 members = [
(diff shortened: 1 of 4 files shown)


langyo:
> Can you make it use `cfg(target_os = "wasi")` instead of a feature image
> 
> That should make the `wasi` feature unnecessary and handle the compilation accordingly. Other than that, it looks good.
> 
> It would be great if you could also update CI to run tests on WASI target (using wasmtime) though

It's so new that when I looked through the documentation, it didn't even mention that the value of `target_os` can be `wasi`...

I tested it on the latest stable compiler and it works. I'll push the modified version later.

hamza1311 (approved):
Thanks! Perhaps this is a good thing to update in the Rust documentation as well

PR #407, opened 2023-11-27
Update http crate to 1.0

dgsantana:
This bumps http to 1.0 and also the gloo-net minor version.


tysen (changes requested):
```toml
rust-version = "1.67"
```

in `Cargo.toml`

ought to resolve these failures.

hamza1311:
I would rather inline the one type that's used from `http` crate than bump the version and MSRV while keeping the requirement of pulling in another dependency. I would be happy to look at any PRs removing the dependency altogether.

dgsantana:
Ok, changed the PR title and inlined the http method code, removing the http dependency.

dgsantana:
Note: The failing and the requirement to increase MSRV is not due to http (MSRV is 1.49) but due to "toml_edit v0.19.15", which in turn is not related to the change in gloo-net.
This is due to proc-macro-crate used in gloo-worker-macros.

dgsantana:
I would propose a new PR to fix this, something like this:
```patch
diff --git a/crates/worker-macros/Cargo.toml b/crates/worker-macros/Cargo.toml
index 474cd46..d560678 100644
--- a/crates/worker-macros/Cargo.toml
+++ b/crates/worker-macros/Cargo.toml
@@ -19,6 +19,8 @@ proc-macro-crate = "1.2.1"
 proc-macro2 = "1.0.47"
 quote = "1.0.21"
 syn = { version = "2.0.15", features = ["full"] }
+# Required to keep msrv at 1.64
+toml_edit = "=0.19.0"
 
 [dev-dependencies]
 trybuild = "1"
```

hamza1311:
The lockfile is present for that. It's most likely that since you updated it before, the lockfile has the previous entry. You can try reverting the change to Cargo.lock and then rebuilding. That _should_ fix the issue

dgsantana:
Done.

crates/net/src/http/method.rs
```diff
@@ -0,0 +1,471 @@
...
+    Put,
+    Delete,
+    Head,
+    Trace,
+    Connect,
+    Patch,
+    // If the extension is short enough, store it inline
+    ExtensionInline(InlineExtension),
+    // Otherwise, allocate it
+    ExtensionAllocated(AllocatedExtension),
```
hamza1311, line 68: Does the browser even support these methods? We don't want to add methods that aren't supposed by the fetch API

crates/net/src/http/method.rs
```diff
@@ -0,0 +1,471 @@
...
+///
+/// ```
+/// use gloo_net::http::Method;
+///
+/// assert_eq!(Method::GET, Method::from_bytes(b"GET").unwrap());
+/// assert!(Method::GET.is_idempotent());
+/// assert_eq!(Method::POST.as_str(), "POST");
+/// ```
+#[derive(Clone, PartialEq, Eq, Hash)]
+pub struct Method(Inner);
```
hamza1311, line 46: I would rather just have `Method` be an enum, unless there's a reason for making this be this way that I'm not aware of?

crates/net/src/http/method.rs
```diff
@@ -0,0 +1,471 @@
...
+
+impl fmt::Display for InvalidMethod {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("invalid HTTP method")
+    }
+}
+
+impl Error for InvalidMethod {}
+
+mod extension {
```
hamza1311, line 309: Is there any use for this with fetch API? If so, what?

hamza1311 (commented):
There are some things in the http crate that may be unnecessary here. Can you ensure we have only what's needed

dgsantana:
Minimum Method version to work with gloo-net.

crates/net/src/http/method.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub const POST: Method = Method::Post;
+    /// The PUT method requests that the enclosed entity be stored under the supplied Request-URI.
+    pub const PUT: Method = Method::Put;
+    /// The DELETE method requests that the origin server delete the resource identified by the Request-URI.
+    pub const DELETE: Method = Method::Delete;
+    /// The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response.
+    pub const HEAD: Method = Method::Head;
+    /// The PATCH method requests that a set of changes described in the
+    /// request entity be applied to the resource identified by the Request-URI.
+    pub const PATCH: Method = Method::Patch;
```
hamza1311, line 59: These aliases just pollute the API, imo. `Method::POST` has no need when `Method::Post` exists

crates/net/src/http/mod.rs
```diff
@@ -14,13 +14,14 @@
...
 mod headers;
+mod method;
 mod query;
 mod request;
 mod response;
 
 pub use headers::Headers;
-#[doc(inline)]
-pub use http::Method;
+// This is http 1.0 version Method inlined in gloo-net so we don't need to depend on http crate
```
hamza1311, line 23: This is no longer true

crates/net/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gloo-net"
-version = "0.4.0"
+version = "0.5.0"
```
hamza1311, line 3: Bumping the version isn't for this PR. That will be done at release

hamza1311 (commented):
Can you also update the CHANGELOG.md accordingly?

tysen:
This mergeable now?

crates/net/src/http/method.rs
```diff
@@ -0,0 +1,86 @@
+use std::{fmt::Display, str::FromStr};
+
+/// HTTP methods that can be used in a fetch request.
+/// The methods as defined by the [fetch spec](https://fetch.spec.whatwg.org/#methods).
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum Method {
```
futursolo, line 6: I prefer how the http crate handled it. We should use a struct with associated constants and not enum as you can choose your own http verb.

See: https://github.com/rustwasm/gloo/pull/312#discussion_r1148362780

hamza1311: I'm not so sure if that's a good idea. Custom HTTP verbs (as far as I know) are not standard. Does fetch even support them?

dgsantana: Just decide what is the better way, I can make the change.

dgsantana:
Ok, I will change the PR to use the http 1.x again, it's simple.

Saber Haj Rabiee ***@***.***> escreveu no dia segunda,
18/12/2023 à(s) 16:23:

tysen:
Maybe update the title of the PR now that `http` has returned?

CHANGELOG.md
```diff
@@ -157,6 +157,7 @@
 
 ### Version "0.5.0"
 
+- Update http crate to 1.0
```
hamza1311, line 160: Can you move this out of v0.5.0? That version was released so this should go in the next one

hamza1311 (commented):
Just one change, then this can be merged. Thanks for your work!

CHANGELOG.md
dgsantana: Of course

dgsantana: I moved to 0.6.0, because changing http to 1.0 is a breaking change.

PS: Happy Holidays everyone and have a awesome time family and friends.

tysen:
This all set now?

CHANGELOG.md
```diff
@@ -155,8 +155,13 @@
 
 ## `net`
 
+### Version "0.6.0"
+
+- Update http crate to 1.0
+
 ### Version "0.5.0"
 
+- Update http crate to 1.0
```
futursolo, line 164: Could you please remove this line from v0.5.0 since it is moved to v0.6.0?
Could you please also add the pull request number like other change log lines?

hamza1311 (approved):
Apologies for the very late review. This looks good. I'll merge it now

torokati44:
A new release with this (and possibly other, currently open, dependency bump PRs) in it would be greatly appreciated! ^^

PR #410, opened 2023-11-30
Add WASI example for CI.

langyo:
#405 

I also made an additional submission, but it was a bit late. Could you please check if this new submission can be merged?

😂


langyo:
> oops, sorry I merged it too quickly. Can you rebase it from master?

Oh..I think I have made some mistake...

I'll create a new PR later and this PR would close...

PR #411, opened 2023-11-30
Add WASI example for CI.

langyo:
#405 

I'm still testing this part of the new testing process, but for some reason the CI testing hasn't been going well...

This PR will remain in draft status until it is ready...

### DIFF
--- a/crates/history/src/any.rs
+++ b/crates/history/src/any.rs
@@ -1,6 +1,8 @@
 use std::borrow::Cow;
 
+#[cfg(not(target_os = "wasi"))]
 use crate::browser::BrowserHistory;
+#[cfg(not(target_os = "wasi"))]
 use crate::hash::HashHistory;
 use crate::history::History;
 use crate::listener::HistoryListener;
@@ -13,8 +15,10 @@ use crate::{error::HistoryResult, query::ToQuery};
 #[derive(Clone, PartialEq, Debug)]
 pub enum AnyHistory {
     /// A Browser History.
+    #[cfg(not(target_os = "wasi"))]
     Browser(BrowserHistory),
     /// A Hash History
+    #[cfg(not(target_os = "wasi"))]
     Hash(HashHistory),
     /// A Memory History
     Memory(MemoryHistory),
@@ -23,7 +27,9 @@ pub enum AnyHistory {
 impl History for AnyHistory {
     fn len(&self) -> usize {
         match self {
+            #[cfg(not(target_os = "wasi"))]
             Self::Browser(m) => m.len(),
+            #[cfg(not(target_os = "wasi"))]
             Self::Hash(m) => m.len(),
             Self::Memory(m) => m.len(),
         }
@@ -31,7 +37,9 @@ impl History for AnyHistory {
 
     fn go(&self, delta: isize) {
         match self {
+            #[cfg(not(target_os = "wasi"))]
             Self::Browser(m) => m.go(delta),
+            #[cfg(not(target_os = "wasi"))]
             Self::Hash(m) => m.go(delta),
             Self::Memory(m) => m.go(delta),
         }
@@ -39,7 +47,9 @@ impl History for AnyHistory {
 
     fn push<'a>(&self, route: impl Into<Cow<'a, str>>) {
         match self {
+            #[cfg(not(target_os = "wasi"))]
             Self::Browser(m) => m.push(route),
+            #[cfg(not(target_os = "wasi"))]
             Self::Hash(m) => m.push(route),
             Self::Memory(m) => m.push(route),
         }
@@ -47,7 +57,9 @@ impl History for AnyHistory {
 
     fn replace<'a>(&self, route: impl Into<Cow<'a, str>>) {
         match self {
+            #[cfg(not(target_os = "wasi"))]
             Self::Browser(m) => m.replace(route),
+            #[cfg(not(target_os = "wasi"))]
             Self::Hash(m) => m.replace(route),
             Self::Memory(m) => m.replace(route),
         }
@@ -58,7 +70,9 @@ impl History for AnyHistory {
         T: 'static,
     {
         match self {
+            #[cfg(not(target_os = "wasi"))]
             Self::Browser(m) => m.push_with_state(route, state),
+            #[cfg(not(target_os = "wasi"))]
             Self::Hash(m) => m.push_with_state(route, state),
             Self::Memory(m) => m.push_with_state(route, state),
         }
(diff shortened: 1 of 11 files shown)


langyo:
cc @hamza1311 😉

Maybe it's time to release a patch version in cargo that supports `yew`'s WASI related compilation targets.

hamza1311:
@langyo gloo-net 0.5 has been released. Can you update this PR?

crates/history/src/any.rs
```diff
@@ -13,8 +15,10 @@ use crate::{error::HistoryResult, query::ToQuery};
 #[derive(Clone, PartialEq, Debug)]
 pub enum AnyHistory {
     /// A Browser History.
+    #[cfg(not(target_os = "wasi"))]
```
futursolo, line 18: What are the effects of enabling BrowserHistory and MemoryHistory for wasi?

I think wasi should have BrowserHistory and MemoryHistory enabled like other targets.
This helps to avoid additional feature flags for downstream crates and will lint the code like other targets.

langyo: > What are the effects of enabling BrowserHistory and MemoryHistory for wasi?
> 
> I think wasi should have BrowserHistory and MemoryHistory enabled like other targets. This helps to avoid additional feature flags for downstream crates and will lint the code like other targets.

If don't, `BrowserHistory` is directly dependent on `wasm-bindgen`. It will cause additional symbols like `__wbindgen_placeholder__::xxx` to appear in the compiled result WASM, which is invalid in WASI.

These changes that I've made to some other repositories recently were also made to fix this.

futursolo: I think WASI has been disabled in https://github.com/rustwasm/wasm-bindgen/pull/3233?

langyo: > I think WASI has been disabled in [rustwasm/wasm-bindgen#3233](https://github.com/rustwasm/wasm-bindgen/pull/3233)?

It seems like it will still try to import symbols if you actively reference it. This is a bit different from the situation where `wasm-bindgen` passively imports utility functions related to heap allocation.

Personally, I feel that it is not a bad thing to proactively limit the compilation targets available to the code. Because this allows the compiler to warn developers faster.

futursolo: > It seems like it will still try to import symbols if you actively reference it.

I think you may need to run `cargo update`. Since this is only released in 0.2.88 last month.

I tried the following code:

```rust
fn main() {
    js_sys::Date::now();
}
```

image


> Personally, I feel that it is not a bad thing to proactively limit the compilation targets available to the code. Because this allows the compiler to warn developers faster.

This is a design decision made all the way up to wasm-bindgen.
One of the reasons is that you can write code targeting both `wasm` (client) and native targets (server) at the same time without having to reconfigure toolchain.

image

langyo: > > It seems like it will still try to import symbols if you actively reference it.
> 
> I think you may need to run `cargo update`. Since this is only released in 0.2.88 last month.
> 
> I tried the following code:
> 
> ```rust
> fn main() {
>     js_sys::Date::now();
> }
> ```
> 
> image
> > Personally, I feel that it is not a bad thing to proactively limit the compilation targets available to the code. Because this allows the compiler to warn developers faster.
> 
> This is a design decision made all the way up to wasm-bindgen. One of the reasons is that you can write code targeting both `wasm` (client) and native targets (server) at the same time without having to reconfigure toolchain.
> 
> image

Oh……

But this update requires updating the CI of all projects compiled using `wasm-bindgen`. I'll still keep [the command line arguments](https://github.com/bytecodealliance/wasmtime/pull/4312) `--trap-unknown-imports` added to the commit's part about calling `wasmtime` CLI.

langyo: Updated on https://github.com/yewstack/yew/pull/3534/commits/66996868523716ef3743ccd24e880c3ba27d17e3

futursolo: But this update requires updating the CI cache of all projects compiled using wasm-bindgen. 

We can restraint `wasm-bindgen = “0.2.88"` for gloo-history, all downstream crates will automatically update to use 0.2.88 or later once they use a gloo-history version that supports wasi.

langyo: > > But this update requires updating the CI cache of all projects compiled using wasm-bindgen.
> 
> We can restraint `wasm-bindgen = "0.2.88"` for gloo-history, all downstream crates will automatically update to use 0.2.88 or later once they use a gloo-history version that supports wasi.

I've create a PR #423 to fix it.

futursolo: @langyo Since 0.2.88 solves the WASI issues, would it be possible to create a pull request to remove the feature flags from gloo-history so we do not subsequently introduce more feature flags into Yew?

langyo: > @langyo Since 0.2.88 solves the WASI issues, would it be possible to create a pull request to remove the feature flags from gloo-history so we do not subsequently introduce more feature flags into Yew?

I think it's better not to remove it yet. Keeping some flags will help users who compile with WASI to quickly locate the cause of the error.

But I will actively consider this matter... Perhaps there's no need to resort to strong measures to circumvent this problem.

futursolo: > Keeping some flags will help users who compile with WASI to quickly locate the cause of the error.

I think it is important to ensure consistency between gloo and wasm-bindgen handling.

When users wrongly uses gloo-history under wasi with client side rendering code, they pretty much uses wasm-bindgen and web-sys wrongly as well. So the user will inevitably be facing runtime errors even if gloo-history uses compiler flags.

I have tried to apply a similar approach with target flags to make Yew `Send`, and it resulted in a very large number of feature flags being used and it will require every user who writes SSR to apply a large number of feature flags.
I have reached a conclusion that using feature flags as unfeasible for end users.

In addition, it will eventually be possible for wasm32-wasi to run under browser environments.
These feature flags will need to be removed to provide that support, which will be another breaking changes.

langyo: Your opinion is also very reasonable. I thought about it for a while and I'll try to roll back a part of the code later.


On the possibility of WASI running in the browser. Personally, I don't think it would be appropriate for WASI to be a well-designed virtual machine interface that would be arbitrarily introduced as a non-standard interface. It's not for technical reasons, but because of some historical lessons... If some non-standard interfaces are adopted on a large scale, it will be very troublesome to standardize for them in the future.

As a recent example, `wasmer` has actually made an interface that is slightly different from standard WASI. Even though this made it become the first WASM runtime written in Rust to be able to communicate over the network, it came at the cost of introducing a large number of specialized patch libraries, even including the Rust compiler.

futursolo: > On the possibility of WASI running in the browser.

Browsers are already on track to support WASI.
What is not yet decided is that whether and how wasm-bindgen wants to support wasi.

https://bugzilla.mozilla.org/show_bug.cgi?id=1701197

PR #420, opened 2023-12-07
Add method to terminate a worker

### LINKED ISSUE #408: Allow calling `terminate` on workers
It may not necessarily be possible to break up the computation inside a Worker into nice chunks (e.g. when calling into an external library to do a long-running computation). Thus I haven't been able to find a nice way to force a Worker to stop (dropping the bridge will eventually destroy the worker once computation is done, but will keep running the task until that point). Would it be possible to add support for hard-terminating a Worker using the [`terminate`](https://rustwasm.github.io/wasm-bindgen/api/web_sys/struct.Worker.html#method.terminate) method?

JonasAlaif:
Adds a `native_worker: RefCell<Option<DedicatedWorker>>` field to `WorkerBridgeInner` to allow immediately terminating a worker. Since there is no way to tell if a worker has been terminated from a `web_sys::Worker`, uses the wrapping `Option` to indicate this.

Also removes some redundant `Rc`. Now there are only two `Rc`-wrapped objects:
- the `WorkerBridgeInner` instance shared between all forked bridges and (weakly) with the message receive handler.
- the callbacks shared between a `WorkerBridge` and (weakly) the `WorkerBridgeInner`.

Fixes #408.


JonasAlaif:
I imagine the CI will be able to run after merging https://github.com/rustwasm/gloo/pull/419

JonasAlaif:
Not sure if I could bump the version within the PR or if that's done later?

hamza1311:
> Not sure if I should bump the version within the PR or if that's done later?

I've merged those changes. You can get the latest changes from master

crates/worker/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gloo-worker"
-version = "0.5.0"
```
hamza1311, line 3: Please don't bump the version here. That'll be done when releasing

hamza1311 (changes requested):
CHANGELOG.md in the root of the repo also needs to be updated. You can add an entry for `latest` version

JonasAlaif:
Awesome, thanks! I've updated the changelog, but I'm not sure if you wanted me to add a new `latest` heading instead.

futursolo (commented):
I am not sure if we should support termination.

Threads in Rust cannot be terminated at all without explicitly having a synchronous mechanism sending signals to them and they exit themselves.
A Worker is technically a thread and I feel it also should not support termination.

Terminating the worker would render the remaining bridges “poisoned”, however the current bridges simply drops any new messages and does not report the termination.

In addition, if `ReactorBridge::next().await` or `OneshotBridge::run().await` is called, they will be blocked forever.

PR #421, opened 2023-12-08
Add option to spawn workers as a module

JonasAlaif:
Workers created with the `--target web` flag must be spawned as a module. This adds the option to do that, alleviating the current `--target no-modules` missing features described [here](https://rustwasm.github.io/wasm-bindgen/examples/without-a-bundler.html?highlight=no-modules#using-the-older---target-no-modules). I will create a linked PR in trunk to add an option to create workers with this flag.


crates/worker/src/actor/spawner.rs
```diff
@@ -18,26 +18,39 @@ use super::traits::Worker;
...
+        .unwrap();
+        let url = Url::create_object_url_with_blob(&blob).unwrap();
+        std::borrow::Cow::Owned(url)
+    };
+    let path = path.as_ref();
+
+    if options.as_module {
+        let mut options = WorkerOptions::new();
+        options.type_(WorkerType::Module);
+        DedicatedWorker::new_with_options(path, &options).ok()
```
futursolo, line 50: When a JavaScript Module is used, a different shim is needed.

```javascript
import init from '{js_shim_url}';
await init();
```

crates/worker/src/actor/spawner.rs
```diff
@@ -176,8 +188,61 @@ where
...
+        let worker = create_worker(options)?;
 
-        self.spawn_inner(worker)
+        Some(self.spawn_inner(worker))
+    }
+}
+
+/// Options for spawning a worker.
+#[derive(Debug, Copy, Clone)]
+pub struct SpawnOptions<'a> {
```
futursolo, line 209: Using options to store this would hinder tree shaking as it would be impossible for the compiler to understand whether a loader / module is used at compile time.

I would prefer to make the functions separate as `spawn`, `spawn_with_loader`, and `spawn_as_module` so compiler can perform tree shaking on the shim script if a loader is used.

If you wish to keep this pattern, it should be moved to spawner types, so a builder pattern can be called like the following:

```rust
MyWorker::spawner()
    .with_loader()
    .as_module()
    .spawn("./my_worker.js");
```

crates/worker/src/actor/spawner.rs
```diff
@@ -176,8 +188,61 @@ where
...
+    /// worker.
+    pub fn with_loader(mut self) -> Self {
+        self.with_loader = true;
+        self
+    }
+
+    /// The worker will be spawned with
+    /// [`options.type`](https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker#type)
+    /// set to `module`. This is currently supported on all browsers except
+    /// Safari (see compatibility table at above link). If it is not supported,
```
futursolo, line 237: I have tested this and seems like Safari does support `{ type: ‘module' }`.

The MDN documentation might be outdated as it indicates Safari as supporting ESModule Worker but not `{type: ‘module’}`. However, it is not possible to create ESModule Worker without supporting `type` first.

crates/worker/src/oneshot/spawner.rs
```diff
@@ -64,4 +65,17 @@ where
 
         OneshotBridge::new(inner, rx)
     }
+
+    /// Spawns a Worker according to the given options. Returns `None` if failed.
+    pub fn spawn_with_options(mut self, options: SpawnOptions<'_>) -> Option<OneshotBridge<N>>
```
futursolo, line 70: To simplify the API, most gloo methods have adopted panicking instead of `Error`.
If a spawning has failed, it should panic like other spawning methods.

There is no guarantee that spawning is going to success even if the API call is successful as spawning process is asynchronous.

futursolo (commented):
Thanks for the pull request.

TBH, I am fine with switching to ESModule-only for the Worker implementation since all major browser has support of it, the ESModule adoption for Web Worker is pretty overdue at this point of time.
However, it is also good to have a non-breaking upgrade path so people are not forced to switch immediately.

crates/worker/src/oneshot/spawner.rs
futursolo: I think if that is the direction we decide to go, then we should update the API properly.
I would prefer dedicated PRs that revisits the API on a per-crate basis and not trying to sneak in with other PRs.

hamza1311: I agree. Panicking is fine for this PR

crates/worker/src/actor/spawner.rs
JonasAlaif: I had no idea about that, so I'm happy to go with whatever you think will make this work. I prefer the latter solution since otherwise if any more options are added the number of `spawn_*` options is exponential (i.e. we'd also need `spawn_with_loader_as_module`)

crates/worker/src/actor/spawner.rs
JonasAlaif: Shouldn't have blindly trusted the MDN doc, will test myself as well and update

crates/worker/src/actor/spawner.rs
JonasAlaif: How does my current implementation look? It's a bit annoying that `as_module` takes the `bool` argument while `with_loader` doesn't: they should probably be consistent.

Pros from having `bool` arg:
- The spawn function takes `&self` and so could be used to spawn multiple times: I could use it as e.g. if I try spawning and get an error (assuming `spawn` doesn't panic) then I toggle some settings and try again.
- I think `as_module` is a nicer name than `as_classic`.

Cons:
- More cumbersome

JonasAlaif:
I've changed the default to spawn the worker as a module (which means that this will be a breaking change), but I think it's worthwhile keeping the old behavior around for backwards compatibility reasons.

crates/worker/src/actor/spawner.rs
futursolo: I think both ways are fine.
However, I think it is important to make sure the API is consistent so that loader and module have same kind of schematics.

JonasAlaif:
Ready to merge now. Will be a breaking change regardless, since both the public API of spawner is changed as well as the default `as_module` is changed.

JonasAlaif:
@hamza1311 or @futursolo if you have time :)

TaoVonQi:
I really need this PR. Any timeline when it will be available?

JonasAlaif:
@ranile any chance you could have a look at merging this? Hopefully there aren't any conflicts in the meantime

LucaCappelletti94:
I also need this PR, any updates?

PR #423, opened 2023-12-09
Cargo lock update

langyo:
cc #411

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
 dependencies = [
  "gimli",
 ]
@@ -46,9 +46,9 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "backtrace"
-version = "0.3.68"
+version = "0.3.69"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4319208da049c43661739c5fade2ba182f09d1dc2299b32298d3a31692b17e12"
+checksum = "2089b7e3f35b9dd2d0ed921ead4f6d318c27680d4a5bd167b3ee120edb105837"
 dependencies = [
  "addr2line",
  "cc",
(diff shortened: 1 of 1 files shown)


futursolo (commented):
Could you please add `wasm-bindgen = "0.2.88"` to `crates/history/Cargo.toml`?

`Cargo.lock` in gloo is not honoured if gloo-history is being referenced as a dependency in a different workspace.

langyo:
> Could you please add `wasm-bindgen = "0.2.88"` to `crates/history/Cargo.toml`?
> 
> `Cargo.lock` in gloo is not honoured if gloo-history is being referenced as a dependency in a different workspace.

Done.

futursolo:
@hamza1311 Do you want to raise MSRV to 1.67?

I am fine with raising MSRV, but this is technically a breaking change, so we might have to raise the minor version.
wasm-bindgen itself does not raise the MSRV, so we can avoid increasing MSRV if we do not update all dependencies.

hamza1311:
I'm fine with bumping the MSRV but it has to be for a good reason. What dependency needs to be bumped here?

If that will push usage of wasm-bindgen `>=0.2.88`, that's a good enough reason for me

futursolo:
> What dependency needs to be bumped here?

Raise of MSRV is side effect of `cargo update`. `wasm-bindgen` does not raise MSRV.
We only need `wasm-bindgen >= 0.2.88`.

langyo:
> I'm fine with bumping the MSRV but it has to be for a good reason. What dependency needs to be bumped here?
> 
> If that will push usage of wasm-bindgen `>=0.2.88`, that's a good enough reason for me

I've checked the CI log and found it finally caused by `gloo-worker-macros`. It depends the new version of `proc-macro-crate`.

hamza1311:
Can you just update the dependencies using `cargo +nightly update -Zmsrv-policy`? That _should_ update the dependencies to MSRV compatible versions. I don't want to raise the MSRV just for a proc-macro dependency

langyo:
> Can you just update the dependencies using `cargo +nightly update -Zmsrv-policy`? That _should_ update the dependencies to MSRV compatible versions. I don't want to raise the MSRV just for a proc-macro dependency, primarily because it forces the same for downstream crates.

Ummm...Sure

I'll try it later

hamza1311:
I created #424 which bumps dependencies without raising MSRV

langyo:
> I created #424 which bumps dependencies without raising MSRV

Okay..

By the way, you reminded me that I'll be reworking the general PR on the `yew` later to change the MSRV part.

futursolo (commented):
@langyo

Could you please remove the MSRV changes and rebase `Cargo.lock` from master?

langyo:
> @langyo
> 
> Could you please remove the MSRV changes and rebase `Cargo.lock` from master?

Oh...I was made a mistake..I rolled back too far...

But since @hamza1311 has completed this part of the work..In fact, this PR is a duplicate, and it should not be a problem to turn it off.

Next I'll see how to remove the compilation macro of wasi...

PR #424, opened 2023-12-09
Update Cargo.lock (2)

hamza1311:
#423 but without raising MSRV

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


futursolo (commented):
#423 also updates `crates/history/Cargo.toml` to include `wasm-bindgen = "0.2.88"`, so it will be respected when used as a dependency.

If you want to keep #423 to update wasm-bindgen, that’s fine with me as well.

futursolo:
In addition, could you please wait until gloo-history removed the wasi compiler flags before releasing the next version?

See: https://github.com/rustwasm/gloo/pull/411#discussion_r1421379894

langyo:
> In addition, could you please wait until gloo-history removed the wasi compiler flags before releasing the next version?
> 
> See: [#411 (comment)](https://github.com/rustwasm/gloo/pull/411#discussion_r1421379894)

If necessary, I'll removing additional WASI judgment tags when this PR has merged.

futursolo (approved):
Let’s merge this and address `Cargo.toml` in #423.

PR #474, opened 2024-05-22
Use direct global binding for `fetch` in place of web_sys fetch bindings

jbms:
This avoids the need to detect whether the module is running in the main browser thread or in a WebWorker.

This matches how `setInterval` and related functions are handled by the gloo-timers crate.

See https://github.com/rustwasm/wasm-bindgen/discussions/3863


littlebenlittle:
Confirmed with #486 that this is a fix for #484!

ranile (approved):
Thanks!

Issue #484, opened 2024-07-17
`http::Request::send` panics in Node with `wasm-pack`


littlebenlittle:
### Describe the Bug

Calls to `http::Request::send` reach the panic condition [here](https://github.com/rustwasm/gloo/blob/dc0e61e7b24fe0f33d66960385909dc0f4195bd9/crates/net/src/http/request.rs#L336) in NodeJS environment.

### Steps to Reproduce

Full crate: 
[gloo_node.tar.gz](https://github.com/user-attachments/files/16273511/gloo_node.tar.gz)

Run:

```sh
wasm-pack build --target nodejs
node index.js
```

Node Version: v21.6.1

Code:

<details>

```rs
extern crate console_error_panic_hook;
extern crate wasm_bindgen;

use std::panic;
use wasm_bindgen::prelude::*;

#[wasm_bindgen]
pub fn main() {
    panic::set_hook(Box::new(console_error_panic_hook::hook));
    wasm_bindgen_futures::spawn_local(async move {
        gloo_net::http::Request::get("localhost:8080")
            .send() // panics
            .await
            .unwrap();
    });
}
```

</details>

### Additional Info

Seems to be the result of undefined `WorkerGlobalScope`. Node v21+ defines `fetch` on the global context ([ref](https://nodejs.org/api/globals.html#fetch)) so something like this might work?

<details>

```rs
pub async fn send(self) -> Result<Response, Error> {
        let request = self.0;
        let global = js_sys::global();
        let maybe_window =
            Reflect::get(&global, &JsValue::from_str("Window")).map_err(js_to_error)?;
        let promise = if !maybe_window.is_undefined() {
            let window = global.dyn_into::<web_sys::Window>().unwrap();
            window.fetch_with_request(&request)
        } else {
            let maybe_worker = Reflect::get(&global, &JsValue::from_str("WorkerGlobalScope"))
                .map_err(js_to_error)?;
            if !maybe_worker.is_undefined() {
                let worker = global.dyn_into::<web_sys::WorkerGlobalScope>().unwrap();
                worker.fetch_with_request(&request)
            } else {
                // panic!("Unsupported JavaScript global context");
                // instead:
                let maybe_fetch = js_sys::Reflect::get(&global, &JsValue::from_str("fetch")).unwrap();
                if maybe_fetch.is_undefined() {
                    panic!("no global fetch()")
                }
                let fetch = maybe_fetch.dyn_into::<js_sys::Function>().unwrap();
[...]

PR #485, opened 2024-07-17
add case for node global.fetch to http::Request::send

littlebenlittle:
Proposed solution to #484


littlebenlittle:
Closed in favor of #474

PR #486, opened 2024-07-18
Add `wasm-pack` tests for `net` to CI

littlebenlittle:
Tests to reproduce #484 

Fixed by ~~#485~~ #474

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -598,7 +598,7 @@ dependencies = [
  "futures-core",
  "futures-sink",
  "gloo-utils 0.1.7",
- "http",
+ "http 0.2.11",
  "js-sys",
  "pin-project",
  "serde",
@@ -619,7 +619,7 @@ dependencies = [
  "futures-io",
  "futures-sink",
  "gloo-utils 0.2.0",
- "http 1.0.0",
+ "http 1.1.0",
  "js-sys",
  "once_cell",
  "pin-project",
@@ -785,7 +785,7 @@ dependencies = [
  "futures-core",
  "futures-sink",
  "futures-util",
- "http",
+ "http 0.2.11",
  "indexmap 2.1.0",
  "slab",
  "tokio",
@@ -814,7 +814,7 @@ dependencies = [
  "base64",
  "bytes",
  "headers-core",
- "http",
+ "http 0.2.11",
  "httpdate",
  "mime",
  "sha1",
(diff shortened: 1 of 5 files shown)


ranile (dismissed):
Please ensure that the tests aren't failing

crates/net/tests/http.rs
```diff
@@ -3,8 +3,6 @@ use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
 use wasm_bindgen_test::*;
 
-wasm_bindgen_test_configure!(run_in_browser);
```
ranile, line 6: This is needed for browser tests, please keep this in (and gate it behind `wasm32-unknown-unknown` target)

crates/net/tests/query.rs
```diff
@@ -1,8 +1,6 @@
 use gloo_net::http::QueryParams;
 use wasm_bindgen_test::*;
 
-wasm_bindgen_test_configure!(run_in_browser);
```
ranile, line 4: same as previous comment

Cargo.lock
ranile, line 1: This is unrelated, please revert

ranile (changes requested):
You'd want to bring this PR up to date with master to get the latest code for CI

crates/net/tests/http.rs
littlebenlittle: Clippy doesn't like `#[cfg(target="wasm32-unkonwn-unknown")]` but it seems to work. Let me know if I misunderstood.

ranile: I think `target_arch` refers to wasm32 in this case (see https://doc.rust-lang.org/reference/conditional-compilation.html#target_arch). That might be why clippy complains. You want to update CI so clippy is also run on wasm32 target.

You can look at how wasi is cfg-gated in other parts of the code base and do the opposite of it here. If it's `target_os = "wasi"` then you may be able to do `not(target_os = "wasi")` to target browser wasm targets

littlebenlittle: So this seems like a quirk of `wasm_bindgen_test`. As far as I can tell there's no way straightforward way to run in both node and browser.

Here's one suggestion: https://github.com/rustwasm/wasm-bindgen/issues/2571#issuecomment-873533119

Added to latest commit.

ranile (approved):
Having to add a feature is sad but this is the best we can really do. Thanks for the PR!

PR #493, opened 2024-10-15
Recognise that connection handling is a developer concern

huntc:
Removes the logic to test for a connection error with Server Sent Events.

My original attraction to SSE was that the browser would handle the connection and re-connect if it can. It turns out, after many years of using SSE(!), I learn that browsers are not consistent in their reconnection policy. For a start, browsers can decide how long to wait before re-connecting.

SSE will re-connect automatically under certain conditions, but again there's no specification in terms of what these conditions should be. For example, an "503 - service unavailable" does not seem to cause a re-connect for Chrome, yet this is often a transient error.

In the end, you're left with implementing your own connection handling anyway. We therefore push the testing of the connecting state back on the developer.

I also clean up a couple of compiler/clippy warnings along the way, and fix the feature flag test for testing the net modules.

### DIFF
--- a/crates/console/src/lib.rs
+++ b/crates/console/src/lib.rs
@@ -25,7 +25,6 @@ mod macros;
 mod timer;
 
 pub use counter::Counter;
-pub use macros::*;
 pub use timer::Timer;
 
 #[doc(hidden)]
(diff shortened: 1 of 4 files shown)


huntc:
Actually, I think the existing logic is just fine having looked inside WebKit. Closing.

PR #500, opened 2025-03-24
Make Request and RequestBuilder implement IntoFuture

ultrabear:
This PR adds IntoFuture implementations (and Future implementing types `RequestFuture` and `RequestBuilderFuture`) for `Request` and `RequestBuilder`

This makes it possible to have smoother usage of the API, with patterns like `Request::get("/foo").await` now being possible (previously `Request::get("/foo").send().await`.

This is my first time handwriting `Future`'s, so they are simple enum based state machines and or just wrap inner futures, I see no reasons they would be noticeably less performant than the previous implementation however.

I realize as I write that it may have been a better idea to make RequestFuture and RequestBuilderFuture public in an unreachable module (not reference-able by userspace or documented publicly), if that change is wanted or to simply make them doc(hidden), or to just leave it be and give them a sentence of docs explaining what they are, I'm willing to do any of the three in a followup commit within this PR

### DIFF
--- a/crates/net/src/http/request.rs
+++ b/crates/net/src/http/request.rs
@@ -4,7 +4,10 @@ use http::Method;
 use js_sys::{ArrayBuffer, Uint8Array};
 use std::convert::{From, TryFrom, TryInto};
 use std::fmt;
+use std::future::{Future, IntoFuture};
+use std::pin::Pin;
 use std::str::FromStr;
+use std::task::{Context, Poll};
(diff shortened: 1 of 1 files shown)


ultrabear:
Some CI tests are failing because they try to grab `wasm32-wasi`, which has been removed. the target was renamed to `wasm32-wasip1`.

ultrabear:
I do not know why the gloo-worker CI tests failed based on my PR, changes were only made in gloo-net and there does not appear to be a dependency chain where gloo-worker imports gloo-net

PR #501, opened 2025-04-10
[RFC]: impl cache storage

wiseaidev:
Hiya (* ^ ω ^)!

I've been tinkering with the browser and noticed that Gloo storage is missing some handy components, like [cache storage](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage). I've also been doing some wild science with Yew and Wasm: [@opensass](https://github.com/opensass).

One thing up for debate: should cache storage be implemented under the service worker or directly in Gloo storage? For now, I think it's best to implement it in the Storage API.

Lemme know what you think!

### DIFF
--- a/crates/storage/src/cache_storage.rs
+++ b/crates/storage/src/cache_storage.rs
@@ -1,0 +1,145 @@
+use crate::AsyncStorage;
+use js_sys::Array;
+use serde::de::Error;
+use serde::{Deserialize, Serialize};
+use serde_json::{Map, Value};
+use std::future::Future;
+use wasm_bindgen::JsCast;
+use wasm_bindgen::UnwrapThrowExt;
+use wasm_bindgen_futures::JsFuture;
+use web_sys::window;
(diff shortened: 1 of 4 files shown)


crates/storage/src/cache_storage.rs
```diff
@@ -0,0 +1,145 @@
...
+    async fn open_cache() -> Result<Cache> {
+        let promise = Self::raw().open("gloo-cache");
+        let cache = JsFuture::from(promise)
+            .await
+            .map_err(|e| StorageError::JsError(js_sys::Error::from(e).into()))?;
+        Ok(Cache::from(cache))
+    }
+
+    async fn all_keys() -> Result<Vec<String>> {
+        let promise = Self::raw().keys();
```
Madoshakalaka: Doesn't this return the names of all caches, as opposed to just `gloo-cache` ?

Madoshakalaka (changes requested):

PR #505, opened 2025-07-21
Update crate thiserror and others 

martinfrances107:
I  like this project ... it is really useful.

In a way this is a field report from a upstream user.

The leptos framework makes extensive use of gloo
I have a application based on leptos ..
 and am tracking down why there are many duplicate versions version of certain crates being built.
( duplicate but with different version numbers ) 

I want to upgrade "thiserror" to a modern version 

This is open source, so for crates I depend on, I am happy to go a bit further and do all the basic maintence 
I can see, So I did the following 

* ran cargo update
* Moved thiserror up into the root workspace so I only have to bump the version to 2.0 once.
* Updated/Fixed deprecated fn names.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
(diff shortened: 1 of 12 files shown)


martinfrances107:
Of the three remaining failure, it looks like firefox is reaching out of the network to preform some ad blocking funciton ... The CORS policy is blocking a "fetch" request for a URL on the internet .. and failing

Hmm not sure how to fix.

Madoshakalaka (approved):
Thanks for the PR. A bit over the places but very good changes, I like the `workspace = true` extractions.
Coincidentally, MSRV 1.82 might seem a bit arbitrary but [it's the actual MSRV](https://github.com/ranile/gloo/pull/532#issue-4120192074) at the moment due to a transitive dependency.

PR #510, opened 2025-08-04
Add WebSocket::buffered_amount

vi:
Inspired by #337.

### DIFF
--- a/crates/net/src/websocket/futures.rs
+++ b/crates/net/src/websocket/futures.rs
@@ -243,6 +243,11 @@ impl WebSocket {
     pub fn protocol(&self) -> String {
         self.ws.protocol()
     }
+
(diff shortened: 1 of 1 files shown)


jin-eld:
@ranile any chance of getting this merged? Would be really great to have it, right now I have to pull in the gloo-net crate into my tree and patch it in order to get this functionality. It is very helpful when sending a lot of data over a WS connection - the browser seems to accept as much as you throw at it, buffering it internally without blocking, so querying the buffered amount seems to be the only way how one can figure out if the network is keeping up or not.

ranile:
@jin-eld thanks for the ping, just merged. I'm not sure if I'll be able to cut a release soon but you should be able to easily pull it in from a git dependency

jin-eld:
awesome, thank you! pulling it from git is good enough

PR #546, opened 2026-04-01
fix(worker): enable spawning web workers from inside web workers

### LINKED ISSUE #511: gloo_worker: how to spawn web worker inside web worker?

Madoshakalaka:
Fixes #511

`WorkerSpawner::create_worker` used `gloo_utils::window()` to resolve relative script URLs, which panics inside a worker context (no `window` object). This PR replaces it with `js_sys::Reflect` reads on `globalThis.location.href`, which is available in both window and dedicated-worker global scopes.

We also removed `gloo-utils` from `gloo-worker` dependencies (only usage was `window()`).

Noteworthily: this is not a breaking change. None of the gloo-worker's public APIs changed. We simply fixed the bug that workers couldn't be spawned from worker contexts. The version in the CHANGELOG requires no update as a new breaking version is being prepared already.

`examples/nested-worker-yew` is added:

nested-router


Madoshakalaka:
oops. accidentally deleted the branch when trying to rename

PR #547, opened 2026-04-01
fix(worker): enable spawning web workers from inside web workers

### LINKED ISSUE #511: gloo_worker: how to spawn web worker inside web worker?

Madoshakalaka:
Fixes #511

`WorkerSpawner::create_worker` used `gloo_utils::window()` to resolve relative script URLs, which panics inside a worker context (no `window` object). This PR replaces it with `js_sys::Reflect` reads on `globalThis.location.href`, which is available in both window and dedicated-worker global scopes.

We also removed `gloo-utils` from `gloo-worker` dependencies (only usage was `window()`).

Noteworthily: this is not a breaking change. None of the gloo-worker's public APIs changed. We simply fixed the bug that workers couldn't be spawned from worker contexts. The version in the CHANGELOG requires no update as a new breaking version is being prepared already.

`examples/nested-worker-yew` is added:

nested-router


examples/nested-worker-yew/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+version = "0.1.0"
+edition = "2021"
+authors = ["Mattuwu"]
+publish = false
+
+[dependencies]
+gloo-worker = { path = "../../crates/worker" }
+serde = { version = "1", features = ["derive"] }
+console_error_panic_hook = "0.1.7"
+yew = { version = "0.21", features = ["csr"] }
```
ranile, line 12: We probably want to update to the latest version later

ranile (approved):
Looks good

examples/nested-worker-yew/Cargo.toml
Madoshakalaka: It's blocked by our MSRV of 1.82 (Yew 0.22+ requires MSRV 1.84)

ranile: Why not raise MSRV? We should switch to 2024 edition too

Madoshakalaka: On one hand, it's very new (Edition 2024/ 1.85 released a year ago)

On the other, I do see adopters like rand 0.10 and some toml deps.

So maybe Yeah. Edition 2024 will open an opportunity to a lot of improvements.

ranile: #553

PR #549, opened 2026-04-02
Prevent panic in `HashHistory::location()` on malformed hash URLs

### LINKED ISSUE #470: [`gloo_history::HashHistory`] Assertion failed when calling `HashHistory`'s `location()`
### Assertion failed when calling `HashHistory`'s `location()`

`HashHistory`'s `location()` asserts that what comes after the '#' in the URL starts with '/'. However the user could delete that '/', and if you then call `location()`, the assertion will trigger, saying "You cannot use relative path with this history type.".

In my specific case, I have a listener that changes the app's state accordingly when the user changes the URL. I do this using `gloo_history::HashHistory`. If the user inputs a URL with the wrong format, I cannot handle it using `HashHistory`.

Currently the only way I found to handle this is to first check the hash manually using `gloo_history::BrowserHistory`:
```rust
let is_bad = !gloo_history::BrowserHistory::new().location().starts_with("#/");
// Avoid using `HashHistory` if `is_bad`
```

#### Option 1: manually

1. Create `gloo_history::HashHistory`
2. Change URL so that it ends with a '#' or contains something other than a '/' after the '#' (e.g. "http://127.0.0.1:8080/#aaa")
3. Call `location()` on `gloo_history::HashHistory`
4. See how an exception gets thrown due to the failed assertion, saying "You cannot use relative path with this history type."

#### Option 2: unit test

1. Open `crates/history/tests/hash_history.rs`
2. Go to the end of `history_works()`
3. Add the following code:
```rust
// Simulate the user changing the URL
window().location().set_hash("a").unwrap();

let location = history.location(); // <-- Exception thrown here
delayed_assert_eq(|| location.path().to_owned(), || "/").await;
```
4. Run the test with wasm-pack: `wasm-pack test --firefox --headless crates/history/ --test hash_history` (or similar)
5. See test fail due to the exception being thrown

### Expected Behavior

Returning a location with empty path, maybe? Or making `location()` return `Option` so that this returns `None`? I'm not sure (suggestion are welcome). Definitely not an exception

### Actual Behavior

Exception thrown

### Additional Context

`HashHistory` tries to prevent this case from happening, but it does so during thread-local initialization, which only happens the first time you create a `HashHistory`:
```rust
// Hash needs to start with #/.
if current_hash.is_empty() || !current_hash.starts_with("#/") {
    let url = HashHistory::get_url();
    url.set_hash("#/");

    browser_history.replace(url.href());
}
```

This code does work as expected, except since the user can change the URL without triggering a page reload, the assumption that fixing the URL hash once will be enough is not actually true.

Not sure it's an ideal fix (since it is kinda lying about the true state of the URL), but the following patch makes the modified unit test I suggest in the "Steps to Reproduce" section ("Option 2: unit test") pass:

```patch
diff --git forkSrcPrefix/crates/history/src/hash.rs forkDstPrefix/crates/history/src/hash.rs
index 792b43a6074bac21d4f7d1b12475460a782a1a77..e050858fc8b4cbb8eb692349cea2ef7b4536618f 100644
--- forkSrcPrefix/crates/history/src/hash.rs
+++ forkDstPrefix/crates/history/src/hash.rs
@@ -196,18 +196,28 @@ impl History for HashHistory {
     fn location(&self) -> Location {
         let inner_loc = self.inner.location();
         // We strip # from hash.
-        let hash_url = inner_loc.hash().chars().skip(1).collect::<String>();
-
-        assert_absolute_path(&hash_url);
-
-        let hash_url = Url::new_with_base(
-            &hash_url,
-            &window()
-                .location()
-                .href()
-                .expect_throw("failed to get location href."),
-        )
-        .expect_throw("failed to get make url");
+        let hash_url = inner_loc.hash().strip_prefix('#').expect("should not possible to not have the location hash start with '#'");
+
+        let hash_url = if !hash_url.starts_with('/') {
+            Url::new_with_base(
+                "",
+                &window()
+                    .location()
+                    .href()
+                    .expect_throw("failed to get location href."),
+            )
+            .expect_throw("failed to get make url")
+        } else {
+            Url::new_with_base(
+                hash_url,
+                &window()
+                    .location()
+                    .href()
+                    .expect_throw("failed to get location href."),
+            )
+            .expect_throw("failed to get make url")
+        };
 
         Location {
             path: hash_url.pathname().into(),

```

If you like this I could try submitting a PR, but I would like to first hear your thoughts. In any case, I feel that if I'm using `HashHistory` and using its API correctly (providing paths starting with '/', etc.), it shouldn't be possible for assertions to fail.

ranile:
`HashHistory::location()` called assert_absolute_path on the URL hash, which panicked if the hash didn't start with #/. Users can edit the URL bar at any time to produce hashes like #aaa or just #, crashing the application on the next location() call.

Replace the `assert_absolute_path` call in `location()` with graceful normalization:

 - If the hash is empty or doesn't start with `/`, prepend `/` (or default to `/` if empty).
 - Log a warning to the browser console so developers can diagnose unexpected routing behavior.
 - Silently correct the URL in the address bar via `replaceState` so it stays in canonical `#/...` form.

The `push`/`replace` methods continue to assert, since those are developer-controlled inputs where a relative path is always a bug.

This approach mirrors how every mainstream JS router library handles the same problem. None of them throw on a malformed hash:

- **`history` v4 / React Router 5** ([source](https://github.com/remix-run/history/blob/v4/modules/createHashHistory.js)): `addLeadingSlash` normalizes on read; `replaceHashPath` auto-corrects the URL on every `hashchange`.
- **`history` v5 / React Router 6** ([source](https://github.com/remix-run/history/blob/dev/packages/history/index.ts)): `parsePath` with destructuring defaults (`pathname = "/"`) silently falls back to root.
- **React Router v7** ([source](https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/router/history.ts)): Explicit `pathname = "/" + pathname` prepend with a code comment explaining why.
- **Vue Router 3** ([source](https://github.com/vuejs/vue-router/blob/dev/src/history/hash.js)): `ensureSlash()` runs on every `hashchange`, prepends `/` if missing, replaces the URL, and aborts the transition.
- **Vue Router 4** ([source](https://github.com/vuejs/router/blob/main/packages/router/src/history/hash.ts)): `createWebHashHistory` delegates to `createWebHistory` with a `#`-based base, unifying URL parsing across modes.

Fixes #470

---

This prior art was researched by Claude and used to implement the fix in this PR

### DIFF
--- a/crates/history/src/hash.rs
+++ b/crates/history/src/hash.rs
@@ -16,7 +16,16 @@ use crate::{error::HistoryResult, query::ToQuery};
 ///
 /// # Panics
 ///
-/// HashHistory does not support relative paths and will panic if routes are not starting with `/`.
+/// The `push` and `replace` family of methods do not support relative paths
+/// and will panic if the provided route does not start with `/`.
+///
+/// # Hash Normalization
+///
+/// If the URL hash is manually edited by the user to a value that does not
+/// start with `#/`, calling `location()` will **not** panic. Instead, it will:
+/// 1. Log a warning to the browser console.
+/// 2. Normalize the hash by prepending `/` if missing.
+/// 3. Silently correct the URL in the address bar via `replaceState`.
 #[derive(Clone, PartialEq)]
 pub struct HashHistory {
     inner: BrowserHistory,
@@ -195,19 +204,29 @@ impl History for HashHistory {
 
     fn location(&self) -> Location {
         let inner_loc = self.inner.location();
-        // We strip # from hash.
-        let hash_url = inner_loc.hash().chars().skip(1).collect::<String>();
 
-        assert_absolute_path(&hash_url);
+        // Strip the leading '#' from the hash.
+        let raw_hash = inner_loc.hash().strip_prefix('#').unwrap_or("").to_string();
+
+        // Normalize: ensure it starts with '/'. Log a warning if it didn't.
+        let needs_correction = raw_hash.is_empty() || !raw_hash.starts_with('/');
+        let normalized = Self::normalize_hash(&raw_hash);
 
         let hash_url = Url::new_with_base(
-            &hash_url,
+            &normalized,
             &window()
                 .location()
                 .href()
                 .expect_throw("failed to get location href."),
         )
-        .expect_throw("failed to get make url");
+        .expect_throw("failed to make url");
+
+        // Auto-correct the URL in the address bar so it stays canonical.
+        if needs_correction {
+            let url = Self::get_url();
+            url.set_hash(&format!("#{normalized}"));
+            self.inner.replace(url.href());
+        }
 
         Location {
             path: hash_url.pathname().into(),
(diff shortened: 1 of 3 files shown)


crates/history/src/hash.rs
```diff
@@ -225,6 +244,33 @@ impl HashHistory {
...
+            );
+            "/".to_string()
+        } else if !raw.starts_with('/') {
+            web_sys::console::warn_1(
+                &format!(
+                    "[gloo_history] HashHistory: URL hash '#{}' does not start with '/'. \
+                     The hash was normalized to '#/{}'. \
+                     Ensure hash-based routes always begin with '#/'.",
+                    raw, raw
+                )
```
Madoshakalaka, line 265: seems good. except can you inline these two `raw`s? clippy doesn't like it.

Madoshakalaka (approved):

PR #562, opened 2026-04-28
fix: propagate unwind safety to user callbacks across all gloo crates

guybedford:
wasm-bindgen 0.2.117+ added `MaybeUnwindSafe` bounds on `Closure::wrap` and `Closure::once` because closures handed to JS are invoked across a `catch_unwind` boundary under `panic = "unwind"`. Every gloo crate that wraps user (or internal) `Fn`/`FnMut`/`FnOnce` callbacks fails to compile under `-Cpanic=unwind` because the `Box::new(callback) as Box<dyn Fn*>` coercion (and the `FnOnce` trait selection) erases the static `UnwindSafe` bound that wasm-bindgen requires.

This breaks every downstream crate that pulls in gloo and tests under `-Cpanic=unwind` — see the failure on https://github.com/MattiasBuelens/wasm-streams/pull/35.

The fix follows the same pattern in every affected crate:

* Where user callbacks flow through the public API (gloo-timers, gloo-events, gloo-render), surface the requirement via a `CallbackUnwindSafe` marker trait. It resolves to `std::panic::UnwindSafe` under `panic = "unwind"` on wasm and to a no-op blanket otherwise. Callers with non-`UnwindSafe` captures must wrap them in `std::panic::AssertUnwindSafe` at the call site, which is where the invariants can actually be reasoned about.

* Where the closures are purely internal (gloo-net websocket/eventsource, gloo-worker actor trampoline, gloo-timers futures wrappers, gloo-render's internal slot), use `Closure::wrap_assert_unwind_safe` / `Closure::once_assert_unwind_safe` under panic=unwind. The unwind-safety assertion is audited per call site: every internal closure either touches lock-free `mpsc::UnboundedSender` pushes, single-shot `Option<Waker>` takes, or `Rc<RefCell<...>>` slots where the borrow is released before invoking nested user code. A panic mid-callback leaves observers seeing only states already reachable by normal operation (a missing wake, a dropped event).

* All under-panic=unwind code paths are cfg-gated, so default `panic = "abort"` builds are byte-for-byte unchanged and remain compatible with any 0.2.x wasm-bindgen. No version-requirement bumps anywhere.

```rs
// before
pub fn new<F>(millis: u32, callback: F) -> Timeout
where F: 'static + FnOnce(),
{
    let closure = Closure::once(callback);
    ...
}
```

```rs
// after
pub fn new<F>(millis: u32, callback: F) -> Timeout
where F: 'static + FnOnce() + CallbackUnwindSafe,
{
    #[cfg(all(target_arch = "wasm32", panic = "unwind"))]
    let closure = Closure::once_assert_unwind_safe(callback);
    #[cfg(not(all(target_arch = "wasm32", panic = "unwind")))]
    let closure = Closure::once(callback);
    ...
}
```

This is not a breaking change. Under `panic = "abort"` (the default) `CallbackUnwindSafe` is a blanket implemented for every `T`, so the bound is invisible. Every existing caller continues to compile unchanged. Under `panic = "unwind"` gloo previously did not compile at all against wasm-bindgen 0.2.117+, so there are no existing panic=unwind callers to break. The change unblocks the configuration; it does not regress it.

The `panic_unwind_build` CI job now exercises every affected crate, not just gloo-timers default-features, so the regression cannot recur silently in any sibling crate. The nightly-pin workaround is also dropped: the upstream LLVM 22 + `panic=unwind` issue (wasm-bindgen#4929) is resolved.

Tested locally:

* `RUSTFLAGS="-Cpanic=unwind" cargo +nightly build -p gloo-timers -p gloo-events -p gloo-render -p gloo-net -p gloo-worker --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind` — passes.
* `cargo build -p gloo-timers --features futures --target wasm32-unknown-unknown` under both panic strategies — passes.
* Default-strategy `cargo check` + `cargo clippy` on all touched crates — clean.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -119,6 +119,43 @@ jobs:
             wasm-pack test --node crates/$x --no-default-features
           done
 
+  panic_unwind_build:
+    name: Build with -Cpanic=unwind
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      # Recent Rust nightlies with LLVM 22 are temporarily broken with
+      # `panic=unwind` (see wasm-bindgen/wasm-bindgen#4929). Pin to a known-good
+      # nightly to avoid flapping.
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly-2026-01-28
+          target: wasm32-unknown-unknown
+          components: rust-src
+
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            target
+          key: cargo-${{ runner.os }}-panic-unwind-${{ hashFiles('**/Cargo.toml') }}
+          restore-keys: |
+            cargo-${{ runner.os }}-panic-unwind-
+            cargo-${{ runner.os }}-
+
+      # Verify gloo-timers (and any other crate that hosts user closures via
+      # `Closure::wrap` / `Closure::once`) still builds cleanly under
+      # `panic = "unwind"` against the new `MaybeUnwindSafe` bounds added in
+      # wasm-bindgen 0.2.117+. Catches downstream regressions like
+      # https://github.com/MattiasBuelens/wasm-streams/pull/35 from creeping in.
+      - name: Build crates with panic=unwind
+        env:
+          RUSTFLAGS: '-Cpanic=unwind'
+        run: |
+          cargo build -p gloo-timers --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
+
   test-history-wasi:
     name: Test gloo-history WASI
     runs-on: ubuntu-latest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -155,6 +155,7 @@ jobs:
           RUSTFLAGS: '-Cpanic=unwind'
         run: |
           cargo build -p gloo-timers --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
+          cargo build -p gloo-timers --features futures --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
 
   test-history-wasi:
     name: Test gloo-history WASI
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -142,23 +142,10 @@ jobs:
             cargo-${{ runner.os }}-panic-unwind-
             cargo-${{ runner.os }}-
 
-      # Verify every gloo crate that wraps user closures via `Closure::wrap`
-      # / `Closure::once` still builds cleanly under `panic = "unwind"`
-      # against the `MaybeUnwindSafe` bounds added in wasm-bindgen 0.2.117+.
-      # Catches downstream regressions like
-      # https://github.com/MattiasBuelens/wasm-streams/pull/35 from creeping
-      # in. The crates listed here are the ones that host wasm-bindgen
-      # closures; the others have no relevant code path.
       - name: Build crates with panic=unwind
         env:
           RUSTFLAGS: '-Cpanic=unwind'
-        run: |
-          cargo build -p gloo-timers --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
-          cargo build -p gloo-timers --features futures --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
-          cargo build -p gloo-events --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
-          cargo build -p gloo-render --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
-          cargo build -p gloo-net --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
-          cargo build -p gloo-worker --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
+        run: cargo build -p gloo-timers -p gloo-events -p gloo-render -p gloo-net -p gloo-worker --all-features --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
 
   test-history-wasi:
     name: Test gloo-history WASI
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -125,12 +125,9 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      # Recent Rust nightlies with LLVM 22 are temporarily broken with
-      # `panic=unwind` (see wasm-bindgen/wasm-bindgen#4929). Pin to a known-good
-      # nightly to avoid flapping.
       - uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: nightly-2026-01-28
+          toolchain: nightly
           target: wasm32-unknown-unknown
           components: rust-src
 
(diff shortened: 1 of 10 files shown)


Madoshakalaka (changes requested):
The `futures` feature is broken now:

```
RUSTFLAGS='-Cpanic=unwind' cargo +nightly check -p gloo-timers --features futures --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
```
fails with 
```
  error[E0277]: the type `UnsafeCell<Option<Waker>>` may contain interior mutability and a reference
   may not be safely transferrable across a catch_unwind boundary
     --> crates/timers/src/future.rs:143:43
     ...
  note: required for `{closure@crates/timers/src/future.rs:143:43: 143:50}`
        to implement `CallbackUnwindSafe`
```

The CI job runs `cargo build -p gloo-timers --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind` with default features only so this went undetected.

guybedford:
I added the fix for the futures case as well and a CI test to check that in future.

We have since landed the wasm-bindgen change that does these stricter assertions which will be going out in wasm-bindgen@0.2.122 - would be great to have this landed by then!

.github/workflows/tests.yml
```diff
@@ -119,6 +119,44 @@ jobs:
...
+          key: cargo-${{ runner.os }}-panic-unwind-${{ hashFiles('**/Cargo.toml') }}
+          restore-keys: |
+            cargo-${{ runner.os }}-panic-unwind-
+            cargo-${{ runner.os }}-
+
+      # Verify gloo-timers (and any other crate that hosts user closures via
+      # `Closure::wrap` / `Closure::once`) still builds cleanly under
+      # `panic = "unwind"` against the new `MaybeUnwindSafe` bounds added in
+      # wasm-bindgen 0.2.117+. Catches downstream regressions like
+      # https://github.com/MattiasBuelens/wasm-streams/pull/35 from creeping in.
```
Madoshakalaka: > and any other crate

doesn't this step only test the gloo-timers crate?

crates/timers/Cargo.toml
```diff
@@ -15,7 +15,7 @@ rust-version = { workspace = true }
 features = ["futures"]
 
 [dependencies]
-wasm-bindgen = "0.2"
+wasm-bindgen = "0.2.117"
```
Madoshakalaka: `wasm-bindgen/src/closure.rs`: `wrap_assert_unwind_safe` landed in 0.2.109, `once_assert_unwind_safe` in 0.2.110. The actual floor needed is 0.2.110

But I don't see the problem of being a bit more forward compatible.

crates/timers/src/callback.rs
```diff
@@ -4,6 +4,32 @@ use js_sys::Function;
...
+impl<T: std::panic::UnwindSafe> CallbackUnwindSafe for T {}
+
+/// Bound carrying the unwind-safety requirement on [`Timeout::new`] /
+/// [`Interval::new`] callbacks.
+///
+/// Under `panic = "unwind"` on wasm the callback is invoked across a
+/// `catch_unwind` boundary inside `wasm_bindgen`, so this resolves to
+/// [`std::panic::UnwindSafe`]. Under any other panic strategy it is a no-op
+/// blanket. Wrap non-`UnwindSafe` captures in [`std::panic::AssertUnwindSafe`]
+/// at the call site.
```
Madoshakalaka: duplicated docs risking drifting away, maybe try
```rs
/// (single doc block here)
#[cfg(all(target_arch = "wasm32", panic = "unwind"))]
pub trait CallbackUnwindSafe: std::panic::UnwindSafe {}
/* impl */

#[doc(hidden)]
#[cfg(not(all(target_arch = "wasm32", panic = "unwind")))]
/* impl */

#[doc(hidden)]
#[cfg(not(all(target_arch = "wasm32", panic = "unwind")))]
pub trait CallbackUnwindSafe {}
/* impl */
```

.github/workflows/tests.yml
```diff
@@ -119,6 +119,44 @@ jobs:
...
+      - uses: actions/checkout@v6
+
+      # Recent Rust nightlies with LLVM 22 are temporarily broken with
+      # `panic=unwind` (see wasm-bindgen/wasm-bindgen#4929). Pin to a known-good
+      # nightly to avoid flapping.
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly-2026-01-28
+          target: wasm32-unknown-unknown
+          components: rust-src
```
Madoshakalaka: is there an issue for this? An unblocking condition for future maintainers would be good

crates/timers/src/future.rs
```diff
@@ -66,9 +78,20 @@ impl TimeoutFuture {
...
+        // `oneshot::Sender` holds an `Arc<Inner<T>>` whose interior is an
+        // `UnsafeCell`-backed lock, so it is not `UnwindSafe`. The assertion
+        // is sound: the closure is `FnOnce`, `tx` is consumed by `send`, and
+        // nothing observes the sender after the callback returns. `rx` takes
+        // the same lock on poll and treats a missing payload as "cancelled".
+        //
+        // The wrapper is unwound by a helper rather than `let _(x) = w;` or
+        // `w.0` so that RFC 2229 disjoint-capture sees the closure capturing
+        // `AssertUnwindSafe<Sender<()>>` (UnwindSafe) and not the inner
+        // `Sender<()>` (!UnwindSafe). See the discussion in PR #563.
```
Madoshakalaka:
```suggestion
        // `Sender<()>` (!UnwindSafe). See the discussion in PR #562.
```

Madoshakalaka (changes requested):
Additionally, I note all sibling crates fail with the flag:

```
RUSTFLAGS='-Cpanic=unwind' cargo +nightly check -p gloo-{events,render,net,worker} --target wasm32-unknown-unknown -Zbuild-std=std,panic_unwind
```

But it's out of the scope for this PR.

crates/timers/Cargo.toml
guybedford: I have guarded all unwind checks with the `panic=unwind` cfg, so that we don't need to bump the version constraint for non-panic-unwind consumers. Panic unwind consumers then only need to use > 0.2.110, which fits the usage model.

.github/workflows/tests.yml
guybedford: Good point, I've gone and expanded this to all the crates.

crates/timers/src/callback.rs
guybedford: Fixed.

crates/timers/src/future.rs
guybedford: Fixed.

.github/workflows/tests.yml
guybedford: Updated: the CI step now builds every crate that hosts wasm-bindgen closures (gloo-timers default + futures, gloo-events, gloo-render, gloo-net, gloo-worker) under panic=unwind, and the misleading parenthetical is gone. Also fixed all four sibling crates as part of this PR — see the new commit `fix(events,render,net,worker)`.

jxs:
Hi, can we have a patch release with this?
it's unfortunately currently blocking https://github.com/libp2p/rust-libp2p/pull/6446#issuecomment-4498266791
Thanks!

